PR #2, opened 2021-08-29
Apply Flutter best practices [Non-Breaking]

hiroshihorie:
Add [flutter_lints](https://pub.dev/packages/flutter_lints) package, and clear the lint warnings.
This PR is intended to be **non-breaking**, but  clearing some lint errors **need breaking changes**, I will open another Draft PR for that.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -32,7 +32,7 @@ jobs:
       - run: flutter pub get
 
       # Check for any formatting issues in the code.
-      - run: flutter format --set-exit-if-changed .
+      - run: flutter format --set-exit-if-changed -l 100 .
 
       # Statically analyze the Dart code for any errors.
       - run: flutter analyze .
--- a/.gitignore
+++ b/.gitignore
@@ -73,3 +73,6 @@ build/
 !**/ios/**/default.mode2v3
 !**/ios/**/default.pbxuser
 !**/ios/**/default.perspectivev3
+
+# VS Code
+.vscode/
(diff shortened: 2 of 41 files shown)


hiroshihorie:
@davidzhao The tests wont pass because I didn't solve the "Prefer using lowerCamelCase for constant names" lint which requires **potentially breaking changes**.

davidzhao:
Ah I see. could we turn off this rule? Some of those we'd probably want to preserve, like `RTCIceTransportPolicyAll`

Do you have ideas around handling the print violations in example? should we ignore that folder entirely?

hiroshihorie:
@davidzhao  IMO that should be an **enum** maybe something like:

```dart
enum RTCIceTransportPolicy {
  all,
  relay,
}

extension RTCIceTransportPolicyExt on RTCIceTransportPolicy {
  String toStringValue() => {
    RTCIceTransportPolicy.all: 'all',
    RTCIceTransportPolicy.relay: 'relay',
  }[this]!;
}
```

But that's potentially breaking, I don't see references to `RTCIceTransportPolicyAll` and `RTCIceTransportPolicyRelay` as of now tho.

davidzhao:
that's fair. go ahead and update to your suggestions. Since we haven't officially released it *and* none of these enums are in the critical path, it's fine to not maintain compatibility.

davidzhao (approved):
looks good!

hiroshihorie:
I find the 80 chars line length too short for `flutter format`, can we set it to **120**?
https://github.com/dart-lang/dart_style/issues/833

`flutter format --set-exit-if-changed -l 120 .`

davidzhao:
I'm ok with relaxing the constraints a bit. But I think 120 is a bit too long. How about 100?

hiroshihorie:
@davidzhao Made minor improvements to the example

PR #4, opened 2021-08-31
Simulcast, Screen sharing & Various improvements

hiroshihorie:
### Simulcast

_To make simulcast work for iOS/Android, WebRTC itself needs patching + need to modify flutter_webrtc._ 

- [x] Web
- [ ] iOS
- [ ] Android

### Next
- [x] Simulcast option for example app
- [x] Fallback to `ConnectOption`'s `TrackPublishOptions` if `options == null` for `publishVideoTrack`
- [x] More code optimizations for dart syntax
- [x] Screen sharing
- [x] Fix a bug that exit and re-entering room causes exception.
- [ ] Better error handling
- [ ] Test Simulcast with real devices
- [ ] Limit what symbols are Exported by `show`
- [ ] Manage AudioSession

### Notes
- PB imports has been prefixed to avoid potential symbol collision such as `Room`


example/lib/src/controls.dart
```diff
@@ -90,7 +90,7 @@ class _ControlsState extends State<Controls> {
     }
 
     try {
-      await track.restartTrack(LocalVideoTrackOptions(position: position));
+      await track.restartTrack(options: LocalVideoTrackOptions(position: position));
```
davidzhao, line 91: what's the trade off with the named parameter? IMO with a single optional param, it seems simpler to make it positional. Though I'm open to alternative views.

example/lib/theme.dart
```diff
@@ -0,0 +1,53 @@
+//
+//
+//
+
+import 'package:flutter/material.dart';
+import 'package:google_fonts/google_fonts.dart';
+
+//
+// Flutter has a color profile issue so colors will look different
```
davidzhao, line 9: this is great to know !

lib/livekit_client.dart
```diff
@@ -20,3 +18,5 @@ export 'src/track/track_publication.dart';
 export 'src/track/local_track_publication.dart';
 export 'src/track/remote_track_publication.dart';
 export 'src/widget/video_track_renderer.dart';
+export 'src/proto/livekit_rtc.pb.dart';
```
davidzhao, line 21: I don't know if we want to export this entire file. likely similar the below. The protobufs are more like internal implementation details.

lib/src/imports.dart
```diff
@@ -0,0 +1,31 @@
...
+
+export 'package:flutter/foundation.dart';
+export 'package:flutter/material.dart';
+export 'package:flutter_webrtc/flutter_webrtc.dart';
+export 'package:logging/logging.dart';
+export 'package:tuple/tuple.dart';
+export 'package:uuid/uuid.dart';
+export 'package:web_socket_channel/web_socket_channel.dart';
+
+export '../livekit_client.dart';
```
davidzhao, line 21: I don't think we'll want to import this. this index file is designed for external packages to use.

lib/src/extensions.dart
```diff
@@ -1,19 +1,33 @@
...
+    ];
+
     return <String, dynamic>{
       // only supports unified plan
       'sdpSemantics': 'unified-plan',
+      if (iceServersMap.isNotEmpty) 'iceServers': iceServersMap,
       if (iceCandidatePoolSize != null) 'iceCandidatePoolSize': iceCandidatePoolSize,
-      'iceServers': iceServersMap,
-      if (iceTransportPolicy != null) 'iceTransportPolicy': iceTransportPolicy,
+      if (iceTransportPolicy != null) 'iceTransportPolicy': iceTransportPolicy!.toStringValue(),
```
davidzhao, line 30: just an observation here. interesting the `!` is still needed even when it's not null

lib/src/_websocket_api.dart
```diff
@@ -1,4 +1,4 @@
-import 'package:web_socket_channel/web_socket_channel.dart';
+import 'imports.dart';
```
davidzhao, line 1: TBH.. I'm not a fan of the uber-imports file.

the main reason is it makes it less clear to the reader which packages this particular file depends on. Perhaps flutter is different.. but with Ruby, it was often a pain to track down the side effects that libraries have on the current file.

lib/src/participant/local_participant.dart
```diff
@@ -33,55 +25,176 @@ class LocalParticipant extends Participant {
...
-      );
-
-      final pub = LocalTrackPublication(trackInfo, track, this);
-      addTrackPublication(pub);
-      notifyListeners();
-
-      return pub;
-    } catch (e) {
-      return Future.error(e);
+    // try {
```
davidzhao, line 28: while running in a debugger.. I've seen cases where uncaught exceptions in an async function was getting flagged. after doing some research, I couldn't figure out what Flutter's expectations were. Do you know if that's right?

lib/src/signal_client.dart
```diff
@@ -48,31 +38,77 @@ class SignalClient {
...
+    bool reconnect = false,
+    bool validate = false,
+  }) {
+    final uri = Uri.parse(uriString);
+
+    return uri.replace(
+      //
+      // It is possible to enforce WSS
+      //
+      scheme: validate ? 'http' : uri.scheme,
```
davidzhao, line 54: this would break if the server is running behind TLS. in validate mode, it'll be https if original url was wss. Otherwise it'll be http / ws

lib/src/signal_client.dart
```diff
@@ -48,31 +38,77 @@ class SignalClient {
...
+      //
+      // Attempt Validation
+      //
+      try {
+        final validateResponse = await http.get(validateUri);
+        if (validateResponse.statusCode != 200) {
+          throw ConnectError(validateResponse.body);
         } else {
-          completer.completeError(ConnectError());
+          throw ConnectError();
```
davidzhao, line 104: so this section is pretty tricky due to the various states it goes through.. I'm not sure why I had used a completer originally (I had changed the implementation here a few times). Hopefully it's no longer needed.

lib/src/track/local_video_track.dart
```diff
@@ -1,66 +1,66 @@
...
+import '../imports.dart';
 
 /// A video track from the local device. Use static methods in this class to create
 /// video tracks.
 class LocalVideoTrack extends VideoTrack {
-  RTCRtpSender? get sender => transceiver?.sender;
+  //
+  // The latest options used for this track
+  //
+  LocalVideoTrackOptions? latestOptions;
```
davidzhao, line 9: what's this used for?

lib/src/track/options.dart
```diff
@@ -1,69 +1,195 @@
...
-
-  Map<String, dynamic> get mediaConstraints {
-    return <String, dynamic>{
-      'mandatory': params.mediaConstraints,
-      'facingMode': position == CameraPosition.front ? 'user' : 'environment',
-    };
-  }
+  //
+  final CameraPosition position;
+  final VideoPreset params;
```
davidzhao, line 7: we shouldn't change this to VideoPreset. A preset is a set of params that we've provided. If users are creating their own, it will no longer be a preset.

lib/src/extensions.dart
hiroshihorie: Yah... not so cool right? 😅

lib/livekit_client.dart
hiroshihorie: I thought so too but `VideoQuality` is referenced from the example app and I cant remove that line, so I might just **show** `VideoQuality` for now.

example/lib/theme.dart
hiroshihorie: Makes life slightly easier ✌🏻

lib/src/signal_client.dart
hiroshihorie: Yup I'll fix that to match wss/https and ws/http

lib/livekit_client.dart
hiroshihorie: for now `VideoQuality` and `TrackType` are exposed.

lib/src/track/options.dart
hiroshihorie: Makes sense!

lib/src/track/local_video_track.dart
hiroshihorie: Since `mediaStreamTrack.getSettings` doesn't work, my goal was to get the actual resolution from the constraints used for `getUserMedia`

hiroshihorie: Still figuring out how to make it more decent tho, I did make a PR for `mediaStreamTrack.getSettings` but it works only for web https://github.com/livekit/flutter-webrtc/pull/3 I'll merge it later.

lib/src/_websocket_api.dart
hiroshihorie: Since you're traumatized with Ruby, I can go back to explicitly importing all files then ✌🏻😁.

lib/src/participant/local_participant.dart
hiroshihorie: Looks good to me but I will need to do more real testing, I found another bug that needs to be fixed also.

lib/src/signal_client.dart
hiroshihorie: After re-checking you were right something was weird so I fixed it, please check next commit. 👍🏻

hiroshihorie: Should be fixed in 73578a497fde6fe1f8ec524eaadd533a23f7622b

example/lib/src/controls.dart
hiroshihorie: Fixed 88d04bfeff745e5217673d7474190863ee28f4a9 !

lib/src/track/options.dart
hiroshihorie: Fixed in b78ded11e976423787def34cf5c4cfabb6928509

hiroshihorie:
I will make SignalClient fire events in serial, not in parallel to make code less prone to unintended errors.

hiroshihorie:
Simulcast works for Web but doesn't work for iOS/Android at the moment, the maybe due to **different SDP being generated** and RID based simulcast is not working with livekit-server? I am not sure why but trying to figure it out.

hiroshihorie:
Simulcast for Web works now by fixing a bug in _flutter_webrtc_

hiroshihorie:
Simulcast doesn't work for Android/iOS, still investigating why.

hiroshihorie:
Implemented screen sharing, now it's possible to easily switch camera/screen by passing `options` to `restartTrack`.
```dart
final options = track.currentOptions.copyWith(
  type: LocalVideoTrackType.display, // Make sure it's display,
  // params: VideoParameters.presetFHD169,
);
await track.restartTrack(options);
```

I might make some more changes tho.

hiroshihorie:
Seems like simulcast is not working for the iOS SDK also? https://github.com/livekit/client-sdk-ios/blob/main/Sources/LiveKit/classes/LocalVideoTrack.swift
WebRTC might need a patch. I will try later.

hiroshihorie:
Ah ok figured it out, simulcast for iOS.

davidzhao:
> Ah ok figured it out, **simulcast for iOS**. So I need to patch `WebRTC` and also modify `flutter_webrtc`. Probably same with Android.

that's great! I assume it's similar to what shiguredo did to WebRTC? let's chat about how to get this change into the [webrtc-sdk](https://github.com/webrtc-sdk) repo

lib/src/signal_client.dart
```diff
@@ -180,42 +240,53 @@ class SignalClient {
...
-      default:
-        log('unsupported message: ' + json.encode(msg));
-    }
+  Future<void> _handleMessage(dynamic message) async {
+    //
+    if (message is! List<int>) return;
+
+    final msg = lk_rtc.SignalResponse.fromBuffer(message);
+
+    await lock.synchronized(() async {
```
davidzhao, line 249: I *think* _handleMessage is already called in a synchronous fashion. though it doesn't hurt to ensure that.

hiroshihorie: I think the delegates are getting fired in the middle of another delegate. So this makes it wait for the delegate to finish before firing the next handler.

lib/src/room.dart
```diff
@@ -332,8 +333,15 @@ class Room extends ChangeNotifier with ParticipantDelegate {
     delegate?.onDataReceived(participant, packet.payload);
   }
 
-  void _onTrackAdded(MediaStreamTrack track, MediaStream? stream, RTCRtpReceiver? receiver) {
+  void _onRemoteMuteChanged(String sid, bool mute) {
+    final track = localParticipant.tracks[sid];
+    //
+    // This will trigger signalClient.sendMuteTrack(sid, mute);
     //
+    track?.muted = mute;
```
hiroshihorie, line 341: This will trigger `sendMuteTrack` back to server, which looks like the way JS SDK is implemented.
Is this intended and correct ?

example/ios/Runner/Info.plist
```diff
@@ -22,6 +22,14 @@
 	<string>$(FLUTTER_BUILD_NUMBER)</string>
 	<key>LSRequiresIPhoneOS</key>
 	<true/>
+	<key>NSCameraUsageDescription</key>
+	<string>$(PRODUCT_NAME) will use camera</string>
+	<key>NSMicrophoneUsageDescription</key>
+	<string>$(PRODUCT_NAME) will use microphone</string>
+	<key>UIBackgroundModes</key>
```
davidzhao, line 29: mind updating the README as well? 🙏

lib/src/participant/local_participant.dart
```diff
@@ -29,107 +35,148 @@ class LocalParticipant extends Participant {
...
+        //
+        // TODO: Get actual video dimensions to compute more accurately
+        // mediaTrack.getConstraints() is not implemented for mobile
+        //
+      } catch (_) {
+        logger.log(Level.WARNING, 'Failed to call `mediaStreamTrack.getSettings()`');
+      }
     }
+
+    logger.info('Compute encodings with resolution: ${width}x${height}, options: ${options}');
```
davidzhao, line 109: would be great to change to `fine`

lib/src/participant/participant.dart
```diff
@@ -168,15 +156,26 @@ class Participant extends ChangeNotifier {
...
-      case TrackType.AUDIO:
-        audioTracks[pub.sid] = pub;
-        break;
-      case TrackType.VIDEO:
-        videoTracks[pub.sid] = pub;
-        break;
-      default:
-      // nothing
-    }
+    // switch (pub.kind) {
```
davidzhao, line 159: probably better to delete these

davidzhao (approved):
solid work dude!

davidzhao:
hmm unsure why tests are failing..

hiroshihorie:
> hmm unsure why tests are failing..

No worries I'll check it out!

PR #8, opened 2021-09-10
Subscriber as primary

hiroshihorie:
Support protocol v3

### Next
- [x] Fix Reconnect
- [ ] Emit relevant events


lib/src/rtc_engine.dart
```diff
@@ -1,28 +1,29 @@
 import 'dart:async';
 
-import 'package:flutter_webrtc/flutter_webrtc.dart';
+import 'package:collection/collection.dart';
+import 'package:flutter/foundation.dart';
+import 'package:flutter_webrtc/flutter_webrtc.dart' as rtc;
+import 'package:livekit_client/src/events.dart';
```
davidzhao, line 6: should change to relative imports

davidzhao (approved):
great work!

PR #9, opened 2021-09-17
Switch to new event system

hiroshihorie:
### Next

- [x] Fix dispose (release resources)
- [ ] Update example to use new events
- [x] Clean up code and docs


lib/src/events.dart
```diff
@@ -1,120 +1,246 @@
...
+/// When a new track is published to room *after* the current participant has
+/// joined. It will not fire for tracks that are already published
+// (Relayed from LiveKitEvent)
+class RoomTrackPublishedEvent extends LiveKitEvent {}
+
+/// A [RemoteParticipant] has unpublished a track
+// (Relayed from LiveKitEvent)
+class RoomTrackUnpublishedEvent extends LiveKitEvent {}
+
+/// A subscribed track is no longer available.
```
davidzhao, line 56: Events look really clean! My only comment is if there's a way to make it more clear which events are fired on Room, and which on Participant. Can we add a line indicating that? i.e. `emitted by Room and Participant` ?

lib/src/track/remote_track_publication.dart
```diff
@@ -28,14 +30,14 @@ class RemoteTrackPublication extends TrackPublication {
...
+  bool get isSubscribed {
     if (_unsubscribed) {
       return false;
     }
-    return super.subscribed;
+    return super.isSubscribed;
   }
 
-  set subscribed(bool val) {
+  set isSubscribed(bool val) {
```
davidzhao, line 40:
```suggestion
setSubscribed(bool val) {
```

nit, but isXXXX usually indicates that it's a getter. feels a bit unusual to call `pub.isSubscribed = false`

davidzhao (approved):
great work!

can we remove the commented out sections for the delegate code?

lib/src/events.dart
hiroshihorie: Yea that's confusing! will clean it up!

PR #13, opened 2021-10-01
Organize events

hiroshihorie:
@davidzhao  Should we directly expose PB types `TrackType`, `VideoQuality` ? or just make a public enum and make that public ?


davidzhao:
I think exposing them is fine. do you have any concerns about surfacing them?

PR #15, opened 2021-10-02
Create android & web plugins

hiroshihorie:
seems like pub.dev needs empty plugins even not used to pass some checks.


davidzhao (approved):
👍

hiroshihorie:
I'll just make sure compiles well in all platforms

PR #18, opened 2021-10-14
Video optimizations

hiroshihorie:
- [x] Use events instead of ChangeNotifier internally for `VideoTrackRenderer`.
- [x] Prepare `VideoTrackRenderer` to detect and report visibility/size for future optimizations.

### DIFF
--- a/lib/src/internal/events.dart
+++ b/lib/src/internal/events.dart
@@ -41,21 +41,21 @@ class EnginePublisherIceStateUpdatedEvent extends EngineIceStateUpdatedEvent {
 }
 
 @internal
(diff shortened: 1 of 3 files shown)


hiroshihorie:
We need to decide how we report that info back to server. I just implemented visiblity/size detection at this point.

davidzhao:
yeah so I think we can do the following:

1. if no renderers are attached to the stream OR none of the attached renderers are visible, set RemoteTrackPublication to paused (we should rename enable/disabled to be pause/resume)
2. determine size of the largest renderer, and call SetTrackQuality() to the server. (we should probably allow that to take width/height in pixels, instead of fixed enums).

PR #27, opened 2021-11-05
Connection quality

hiroshihorie:


lib/src/proto/livekit_rtc.pbjson.dart
```diff
@@ -213,6 +213,15 @@ const SignalResponse$json = const {
       '9': 0,
       '10': 'roomUpdate'
     },
+    const {
```
boks1971, line 216: Wow, a bunch of magic numbers and strings. What are these?

boks1971 (approved):
@hiroshihorie looks fine to me, but there is a lot of stuff I do not understand in dart and there are some magic stuff in there. Needs another pair of :eyes: :-)

davidzhao (approved):
lgtm!

lib/src/proto/livekit_rtc.pbjson.dart
davidzhao: these are generated protobuf files :)

hiroshihorie:
Thanks Oniichans 🙏

PR #28, opened 2021-11-08
Simpler APIs

hiroshihorie:


lib/src/types.dart
```diff
@@ -29,6 +29,13 @@ enum Reliability {
   lossy,
 }
 
+enum TrackSource {
+  unknown,
+  camera,
+  microphone,
+  screenShare,
```
davidzhao, line 36: FYI: I also added a ScreenShareAudio type.. for when the screen share is shared with an audio track.

davidzhao (commented):
lg, just a minor comment

lib/src/participant/local_participant.dart
```diff
@@ -230,3 +232,46 @@ class LocalParticipant extends Participant {
     super.updateFromInfo(info);
   }
 }
+
+extension LocalParticipantTrackSourceExt on LocalParticipant {
+  Future<void> setCamercaEnabled(bool enabled) async {
```
davidliu, line 237: nit: typo in `setCamercaEnabled` (should be `setCameraEnabled`)

davidliu (approved):

lib/src/participant/local_participant.dart
hiroshihorie: Omg yesss thanks onichan!!! 🙏🏻😁

PR #33, opened 2021-11-22
macOS support

hiroshihorie:
add macOS support & make example compile


hiroshihorie:
Actually there's nothing much to see so I might go ahead and merge it in 😅🙏

macos/Classes/LiveKitPlugin.swift
```diff
@@ -0,0 +1 @@
+public class LiveKitPlugin: SwiftLiveKitPlugin {}
```
davidliu, line 1: Should the ios `LiveKitPlugin.h/m` files also convert to this?

davidliu (approved):
LGTM

macos/Classes/LiveKitPlugin.swift
hiroshihorie: for some reason Flutter/iOS needs the ObjC version, i know its ugly (inconsistent) 😅

hiroshihorie:
thanks for the review onichan 😁🙏

macos/Classes/LiveKitPlugin.swift
hiroshihorie: actually I'll take a look if it will work Swift only! thanks 😁

Issue #26, opened 2021-11-05
Full HD resoution


Wallacy:
Anyone was able to use this lib using more than 720p resolution? I tried in several ways to setup a 1080p video call but the system never able to use more than 1280x720@30.

Tested using Android, iPhone and Web.

PR #35, opened 2021-11-26
Use min for width/height to apply constraints

davidliu:
Fixes https://github.com/livekit/client-sdk-flutter/issues/26

Turns out that the underlying flutter webrtc only looks at `minHeight` and `minWidth` for applying constraints, and otherwise defaults to 720p.

That said, the library will pop an OverconstrainedError if the requested constraints are too high (at least on Chrome, Android I think will automatically find the closest supported format, haven't checked on iOS). Might need to find a way to make the available formats visible to the user or handle that ourselves.


boks1971 (approved):
Strange one, good catch @davidliu

hiroshihorie:
nice catch!! thanks!

davidzhao:
well done!

PR #41, opened 2021-11-30
Send video dimensions to the server

### LINKED ISSUE #38: publication.dimension not working
when I print out the publication.dimension for every participant, it always returns width:0 height:0 
I've tried it on web and android, both have the same behavior.

I think it used to work fine in previous versions.

testing it with:
livekit-server 0.14.2
client-sdk-flutter 0.5.4

davidzhao:
Fixes #38

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,11 @@ More Docs and guides are available at [https://docs.livekit.io](https://docs.liv
 
 🔴 = Not currently available (Possibly in the future)
 
+
+## Example app
+
+We built a multi-user conferencing app as an example in the [example/](example/) folder. You can join the same room from any supported LiveKit clients.
+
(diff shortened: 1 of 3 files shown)


hiroshihorie (approved):
✅

hiroshihorie:
Swift SDK can detect dimensions updates without relying on the LiveKit Server.
If the LiveKit server uses this information for video optimization purpose, it's better to have an API to report the dimensions change?

davidzhao:
> Swift SDK can detect dimensions updates without relying on the LiveKit Server. If the LiveKit server uses this information for video optimization purpose, it's better to have an API to report the dimensions change?

agreed, we should have an API to update the server when it changes.

PR #43, opened 2021-12-01
Type-aware Publication etc (static typing)

hiroshihorie:
No casting is required anymore for `TrackPublication.track`

`LocalParticipant.videoTracks` returns `List<LocalTrackPublication<LocalVideoTrack>>`


davidzhao (approved):
that's pretty cool! a clever way to produce cleaner code. maybe we should do this across the board.

PR #45, opened 2021-12-02
Option to unpublish without stopping track

hiroshihorie:
Now there is a global option and also overridable at `unpublishTrack` and `unpublishAllTracks`

https://linear.app/livekit/issue/LK-308/api-design-unpublishleave-room-without-stopping-local-tracks

Does it make sense for Local tracks only ?


hiroshihorie:
Would appreciate your thoughts about this design @davidzhao @davidliu 🙂

davidliu:
Seems like it'd only be for local tracks right? I can't think of cases where the track is unpublished, but the track can still technically be subscribed to and receive any track data.

hiroshihorie:
@davidliu Yes I think it makes sense only for Local tracks 😂

lib/src/participant/local_participant.dart
```diff
@@ -170,7 +170,11 @@ class LocalParticipant extends Participant {
 
   /// Unpublish a track that's already published
   @override
-  Future<void> unpublishTrack(String trackSid, {bool notify = true}) async {
+  Future<void> unpublishTrack(
+    String trackSid, {
+    bool notify = true,
+    bool? stopTrack,
```
davidzhao, line 176: I wonder if we should offer both defaults *and* a parameter for each unpublish. what if we didn't have a room option; would it change how users would use this?

lib/src/participant/remote_participant.dart
```diff
@@ -175,7 +175,11 @@ class RemoteParticipant extends Participant {
   }
 
   @override
-  Future<void> unpublishTrack(String trackSid, {bool notify = true}) async {
+  Future<void> unpublishTrack(
+    String trackSid, {
+    bool notify = true,
+    bool? stopTrack,
```
davidzhao, line 181: as DL mentioned, I don't think this makes sense for remote. ideally this is exposed only to the package and not outside of the library.

hiroshihorie:
I have reverted changes, now no option on `unpublishTrack`
engine.connectOptions is never null so there will always be default option.

davidzhao (approved):
looks good!

PR #46, opened 2021-12-02
re/livekit_client/livekit

ashellunts:


hiroshihorie:
We cannot change the package name it's already published 🙂
https://pub.dev/packages/livekit_client

ashellunts:
In that case I think we can't rename `livekit_client_plugin` to `livekit_plugin`.

hiroshihorie:
Ok no problem then 🙂 as long as there are no compile issues.

hiroshihorie:
Thanks for trying to make it consistent with other plugin names 🙂

PR #47, opened 2021-12-06
Fix mute/unmute on Windows

ashellunts:
ReplaceTrack is not implemented in flutter_webrtc.


hiroshihorie:
This will break other platforms can you wrap it with windows platform check? 😁

ashellunts:
> This will break other platforms can you wrap it with windows platform check? 😁

Done

hiroshihorie (approved):
✅ lgtm, but the root issue is that replaceTrack is not implemented yet in Windows ?

ashellunts:
> ✅ lgtm, but the root issue is that replaceTrack is not implemented yet in Windows ?

Yes. We can implement it later.

PR #50, opened 2021-12-07
Fix rendering of remote shared screen

ashellunts:
Fix activeVideoTrack to return any unmuted stream, not only first one in the list.


hiroshihorie (approved):
✅

PR #51, opened 2021-12-09
Default capture options

hiroshihorie:
LK-319
Simply pass default capture settings.
Same as https://github.com/livekit/client-sdk-ios/pull/32

### DIFF
--- a/lib/src/participant/local_participant.dart
+++ b/lib/src/participant/local_participant.dart
@@ -297,7 +297,7 @@ class LocalParticipant extends Participant<LocalTrackPublication> {
     } else if (enabled) {
       if (source == TrackSource.camera) {
         final track = await LocalVideoTrack.createCameraTrack(
-            room.roomOptions?.defaultVideoCaptureOptions);
+            room.roomOptions?.defaultCameraCaptureOptions);
         return await publishVideoTrack(track);
       } else if (source == TrackSource.microphone) {
         final track = await LocalAudioTrack.create(
@@ -305,7 +305,7 @@ class LocalParticipant extends Participant<LocalTrackPublication> {
         return await publishAudioTrack(track);
       } else if (source == TrackSource.screenShareVideo) {
         final track = await LocalVideoTrack.createScreenShareTrack(
-            room.roomOptions?.defaultVideoCaptureOptions);
+            room.roomOptions?.defaultScreenShareCaptureOptions);
         return await publishVideoTrack(track);
       }
     }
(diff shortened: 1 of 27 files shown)


lib/src/participant/local_participant.dart
```diff
@@ -296,13 +296,16 @@ class LocalParticipant extends Participant<LocalTrackPublication> {
...
         return await publishVideoTrack(track);
       } else if (source == TrackSource.microphone) {
-        final track = await LocalAudioTrack.create();
+        final track = await LocalAudioTrack.create(
+            room.roomOptions?.defaultAudioCaptureOptions);
         return await publishAudioTrack(track);
       } else if (source == TrackSource.screenShareVideo) {
-        final track = await LocalVideoTrack.createScreenShareTrack();
+        final track = await LocalVideoTrack.createScreenShareTrack(
+            room.roomOptions?.defaultVideoCaptureOptions);
```
boks1971, line 308: Are these the same in Flutter case? I am guessing we are doing Simulcast for camera and non-Simulcast for screen share. Can Flutter do that?

lib/src/track/local/video.dart
```diff
@@ -49,9 +53,13 @@ class LocalVideoTrack extends LocalTrack with VideoTrack {
...
   static Future<LocalVideoTrack> createScreenShareTrack([
-    ScreenShareTrackOptions? options,
+    VideoCaptureOptions? options,
   ]) async {
-    options ??= const ScreenShareTrackOptions();
+    if (options is VideoCaptureOptions &&
+        options is! ScreenShareCaptureOptions) {
+      options = ScreenShareCaptureOptions.from(captureOptions: options);
+    }
+    options ??= const ScreenShareCaptureOptions();
```
boks1971, line 62: Oh looks like this is overriding default options if it is not screen share options. Seems a bit confusing. Is it possible to just set the defaults to the right values in the caller?

lib/src/participant/local_participant.dart
hiroshihorie: Makes sense to separate defaults for cam and screen share 😁

hiroshihorie:
Sounds like splitting default options for Camera and Screen share is better 🙂

hiroshihorie:
Separate camera & screen share capture options 🙂

lib/src/track/local/video.dart
hiroshihorie: I'm trying to keep it consistent with this pattern 👉 https://github.com/livekit/client-sdk-flutter/blob/main/lib/src/participant/local_participant.dart#L88-L100

For now .. but I think we should move the create track methods to
`LocalParticipant.createCameraTrack` so there is access to the default room options.

davidliu (approved):
LGTM

Issue #57, opened 2021-12-15
android: example app doesn't show remote participants video track


alex-bluetrain:
when running the example app in android, the remote participants can be heard but not seen.
the remote video track appears as if it was muted.

even if the remote participant mutes/unmutes his video or switches camera, it's never seen on the android app.

btw, the chrome version of the app can see the android video stream without problem.

this was tested on 
client-sdk-flutter commit 28e7091325f6a692948037940065d02187a1dbb1
livekit-server commit c00d799ac6ebcfbdf32b48c886801f968a4e62c6

hiroshihorie:
I found why this is happening, thinking what's the best solution now.

PR #58, opened 2021-12-15
Fix: move `waitFor<EngineConnectedEvent>` after creating remote participants

hiroshihorie:
This fixes Issue https://github.com/livekit/client-sdk-flutter/issues/57

The main cause of this issue #57
was waitFor was before creating remote participants from join response.
Which probably happened in the “Create room before connect” PR.

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -120,12 +120,7 @@ class Room extends DisposableChangeNotifier with EventsEmittable<RoomEvent> {
     for (final info in joinResponse.otherParticipants) {
       logger.fine('DEBUG_01 participant: ${info.sid}(${info.identity}) '
           'tracks:${info.tracks.map((e) => e.sid)}');
-      final isNew = !_participants.containsKey(info.sid);
-      final participant = _getOrCreateRemoteParticipant(info.sid, info);
-      if (!isNew) {
-        // update participant in cases where track arrived first and participant was already created.
-        await participant.updateFromInfo(info);
-      }
+      _getOrCreateRemoteParticipant(info.sid, info);
     }
 
     logger.fine('Waiting to engine connect...');
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -123,7 +123,6 @@ class Room extends DisposableChangeNotifier with EventsEmittable<RoomEvent> {
       final isNew = !_participants.containsKey(info.sid);
       final participant = _getOrCreateRemoteParticipant(info.sid, info);
       if (!isNew) {
-        participant.hasInfo;
         // update participant in cases where track arrived first and participant was already created.
         await participant.updateFromInfo(info);
       }
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -114,12 +114,6 @@ class Room extends DisposableChangeNotifier with EventsEmittable<RoomEvent> {
 
     logger.fine('Waiting to engine connect...');
 
-    // wait until engine is connected
-    await _engineListener.waitFor<EngineConnectedEvent>(
-      duration: Timeouts.connection,
-      onTimeout: () => throw ConnectException(),
-    );
-
     localParticipant = LocalParticipant(
       room: this,
       info: joinResponse.participant,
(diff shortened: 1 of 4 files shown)


lib/src/room.dart
```diff
@@ -125,13 +125,20 @@ class Room extends DisposableChangeNotifier with EventsEmittable<RoomEvent> {
...
     for (final info in joinResponse.otherParticipants) {
       logger.fine(
           'DEBUG_01 Existing Participant: ${info.sid}(${info.identity}) tracks:${info.tracks.map((e) => e.sid)}');
-      _getOrCreateRemoteParticipant(info.sid, info);
+      final isNew = !_participants.containsKey(info.sid);
+      final participant = _getOrCreateRemoteParticipant(info.sid, info);
+      if (!isNew) {
+        participant.hasInfo;
+        // update participant in cases where track arrived first and participant was already created.
+        await participant.updateFromInfo(info);
```
hiroshihorie, line 138: we need to update it here

lib/src/room.dart
```diff
@@ -125,13 +125,20 @@ class Room extends DisposableChangeNotifier with EventsEmittable<RoomEvent> {
...
+    // logger.fine('DEBUG_01 Did connect');
 
     for (final info in joinResponse.otherParticipants) {
       logger.fine(
           'DEBUG_01 Existing Participant: ${info.sid}(${info.identity}) tracks:${info.tracks.map((e) => e.sid)}');
-      _getOrCreateRemoteParticipant(info.sid, info);
+      final isNew = !_participants.containsKey(info.sid);
+      final participant = _getOrCreateRemoteParticipant(info.sid, info);
+      if (!isNew) {
+        participant.hasInfo;
```
davidliu, line 136: does this line do anything?

davidliu (commented):
Still a little unclear here; is the track publish event arriving from the server before the new participant join event arrives?

davidzhao:
I just want to note that for the metadata to arrive *after* onTrack could happen, but is very unlikely. Does it explain why Android seems to be the only one having issues receiving remote video tracks? (but audio would work).

hiroshihorie:
The main issue is the waiting for **ice connect logic** was _before_ creating remote participants. Maybe in this PR https://github.com/livekit/client-sdk-flutter/commit/0f5a758cd4ff5e2e8bf12c4a5536afe058d37c1b

Now the order should be correct in this commit https://github.com/livekit/client-sdk-flutter/pull/58/commits/af9957dea304da0f63467dba81ad9bacec9dc9e4

lib/src/room.dart
hiroshihorie: oops! thats not needed 😂

lib/src/room.dart
```diff
@@ -112,6 +112,22 @@ class Room extends DisposableChangeNotifier with EventsEmittable<RoomEvent> {
...
+    );
+
+    for (final info in joinResponse.otherParticipants) {
+      logger.fine('DEBUG_01 participant: ${info.sid}(${info.identity}) '
+          'tracks:${info.tracks.map((e) => e.sid)}');
+      final isNew = !_participants.containsKey(info.sid);
+      final participant = _getOrCreateRemoteParticipant(info.sid, info);
+      if (!isNew) {
+        // update participant in cases where track arrived first and participant was already created.
+        await participant.updateFromInfo(info);
```
hiroshihorie, line 127: I am pretty sure it's very unlikely but I wonder if it's 0 possibility that between
**after receiving the join response** and **creating existing remote participants**
tracks could be received(EngineTrackAddedEvent).

I guess this code is not required and can be reverted.

hiroshihorie:
@davidzhao @davidliu 
The main cause of this issue https://github.com/livekit/client-sdk-flutter/issues/57
was waitFor<IceConnect> was before creating remote participants from join response.
Which probably happened in the “Create room before connect” PR.

Reverted additional checks since it is very unlikely to be required.
Sorry for the confusion 😅

lib/src/room.dart
```diff
@@ -112,24 +112,14 @@ class Room extends DisposableChangeNotifier with EventsEmittable<RoomEvent> {
...
       room: this,
       info: joinResponse.participant,
     );
 
-    // logger.fine('DEBUG_01 Did connect');
-
     for (final info in joinResponse.otherParticipants) {
-      logger.fine(
-          'DEBUG_01 Existing Participant: ${info.sid}(${info.identity}) tracks:${info.tracks.map((e) => e.sid)}');
+      logger.fine('DEBUG_01 participant: ${info.sid}(${info.identity}) '
```
davidzhao, line 121: note on DEBUG_01 logging

davidzhao (approved):

PR #62, opened 2021-12-21
Dynacast

hiroshihorie:
LK-352

For non-simulcast, what is the desired behavior ?

### DIFF
--- a/lib/src/publication/local.dart
+++ b/lib/src/publication/local.dart
@@ -1,5 +1,11 @@
+import 'package:collection/collection.dart';
+import 'package:meta/meta.dart';
+
+import '../extensions.dart';
+import '../logger.dart';
 import '../participant/local.dart';
 import '../proto/livekit_models.pb.dart' as lk_models;
+import '../proto/livekit_rtc.pb.dart' as lk_rtc;
 import '../track/local/local.dart';
 import 'track_publication.dart';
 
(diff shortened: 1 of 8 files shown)


lib/src/publication/local.dart
```diff
@@ -27,4 +33,39 @@ class LocalTrackPublication<T extends LocalTrack> extends TrackPublication<T> {
...
+    if (encodings == null) return;
+
+    bool didChange = false;
+
+    for (final encoding in encodings) {
+      final layer = layers.firstWhereOrNull((e) {
+        // If there is exact match, use it
+        if (e.quality.toRid() == encoding.rid) return true;
+        // Use first layer found if not simulcast
+        if (encoding.rid == null && encodings.length == 1) return true;
```
davidzhao, line 53: when RID isn't defined, we should specifically match it to LOW quality layer. The protocol doesn't guarantee layers to come in in any specific order. (even tho low comes in first).

lib/src/publication/local.dart
```diff
@@ -27,4 +33,39 @@ class LocalTrackPublication<T extends LocalTrack> extends TrackPublication<T> {
...
+        return false;
+      });
+      if (layer != null && encoding.active != layer.enabled) {
+        encoding.active = layer.enabled;
```
davidzhao, line 57: since Flutter includes web, it'll likely suffer from the same issue as web SDK with FireFox. I think it'd be good to put a comment in here, but it's fine if we don't handle it right now.

davidzhao (commented):
this looks good! Question, does this work across platforms with flutter-webrtc translating the RTPSender params?

lib/src/publication/local.dart
boks1971: Is flutter-webrtc using Firefox's WebRTC library for web?

boks1971 (approved):

lib/src/publication/local.dart
davidzhao: they simply translate into equivalent calls from dart to JS. So any browser specific implementation quirks will still remain

hiroshihorie:
> this looks good! Question, does this work across platforms with flutter-webrtc translating the RTPSender params?

I hope it does a good job translating the params but if it doesn't we need to make a PR...

Issue #63, opened 2021-12-24
random video freeze when participant disconnects


alex-bluetrain:
## description

when a participant leaves the room, some of the remaining participant may randomly experience a freeze on the video track from a remote participant or from his own stream.

## how to reproduce

run the client-sdk-flutter/example on three devices (I tried iOS, Android and chrome)
join the same room, then make one participant leave. Then check if all video streams are updating on the remaining participants

if it doesn't occur, rejoin the room on the participant that left, and try again... it will eventually happen.

btw, the audio is still alive, but the video is frozen, it gets fixed if the "frozen" participant mutes/unmutes his video

I think it's related to the visibility events being fired

tested on commit 2b375033e9cb2cd1b7eba2e547248b3b35d3c9d7
livekit-server 0.15.1

alex-bluetrain:
oh, I forgot to mention, sometimes I get the freeze without even leaving the room, but I couldn't find a repro for that scenario, it's probably the same bug.

alex-bluetrain:
as @davidzhao  suggested, I tried disabling the optimizeVideo option on the room, and I could no longer reproduce the issue, it must be something related to it

alex-bluetrain:
I've been testing it for a while without the optimizeVideo, and it's much better, not only the freezing, but also there was some lag induced by the optimizeVideo feature, I had like 1 second video lag when it kicked in, and unnecesary low resolution switching, now the video plays smoothly with no delays and no freezes.

maybe that feature it should be marked as experimental and disabled by default until it's fixed and tested throughly

hiroshihorie:
Thanks alex for the report, I'm looking into this now.

hiroshihorie:
Hi alex this should be fixed in https://github.com/livekit/client-sdk-flutter/pull/65 but I will do more testing to make sure.

hiroshihorie:
This feature has been renamed to `adaptiveStream` in `RoomOptions` and now is **off** by default.
The logic has been re-written and after testing I believe it is much more stable compared to before so I will close this issue for now.

PR #65, opened 2021-12-28
Fix: Visibility Optimization randomly not working

hiroshihorie:
Should fix issue https://github.com/livekit/client-sdk-flutter/issues/63

Seems to work well but further testing is required.

### DIFF
--- a/lib/src/publication/remote.dart
+++ b/lib/src/publication/remote.dart
@@ -172,7 +172,7 @@ class RemoteTrackPublication<T extends RemoteTrack>
           (_) => _computeVideoViewVisibility(),
         );
 
-        newValue.onVideoViewDidBuild = (_) {
+        newValue.onVideoViewBuild = (_) {
           logger.fine('[Visibility] VideoView did build');
           if (_lastSentTrackSettings?.disabled == true) {
             // quick enable
--- a/lib/src/publication/remote.dart
+++ b/lib/src/publication/remote.dart
@@ -104,10 +104,8 @@ class RemoteTrackPublication<T extends RemoteTrack>
     );
 
     // filter visible build contexts
-    final ctxs = videoTrack.viewKeys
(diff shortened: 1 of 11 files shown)


davidzhao:
this looks fine, but I'm not really understanding what caused it to fail before. in which scenarios would it fail before?

davidzhao (approved):

lib/src/publication/remote.dart
```diff
@@ -97,6 +98,10 @@ class RemoteTrackPublication<T extends RemoteTrack>
   // from one of the renderers
   void _onVideoRendererVisibilityUpdateEvent(
       TrackVisibilityUpdatedEvent event) {
+    if (event.track is LocalVideoTrack) {
```
davidzhao, line 101: question: why is the RemoteTrackPublication handling events from a LocalVideoTrack?

hiroshihorie: Right! this should never happen so the check is not necessary. I will remove it.

hiroshihorie:
> this looks fine, but I'm not really understanding what caused it to fail before. in which scenarios would it fail before?
It was unstable because Flutter frequently recreates the widget instances and I think the only reliable way to identify a widget is by using `Key`.
But i think i will just use an alternative (simple) method that doesn't rely on widget ids.

hiroshihorie:
re-designed and simplified so doesn't rely on `visibility_detector`, doesn't rely on view ids...

hiroshihorie:
> this looks fine, but I'm not really understanding what caused it to fail before. in which scenarios would it fail before?


It failed before because the id i used to manage VideoViews were unreliable. using objectId doesn't work because Widget gets rebuilt and also State gets reused.

New approach simply scans all visible VideoViews periodically.

davidzhao:
got it. nice work!

PR #67, opened 2022-01-04
Wait for data channel to `open`

hiroshihorie:
In addition to publisher ice-connect, 
this PR makes it wait until data channel is `open` state.

### DIFF
--- a/lib/src/core/engine.dart
+++ b/lib/src/core/engine.dart
@@ -43,16 +43,11 @@ class Engine extends Disposable with EventsEmittable<EngineEvent> {
   PCTransport? get primary => _subscriberPrimary ? subscriber : publisher;
 
   // data channels for packets
-  rtc.RTCDataChannel? _reliableDC;
-  rtc.RTCDataChannel? _lossyDC;
+  rtc.RTCDataChannel? _reliableDCPub;
+  rtc.RTCDataChannel? _lossyDCPub;
   rtc.RTCDataChannel? _reliableDCSub;
   rtc.RTCDataChannel? _lossyDCSub;
 
-  rtc.RTCDataChannelState get reliableDataChannelState =>
-      _reliableDC?.state ?? rtc.RTCDataChannelState.RTCDataChannelClosed;
-
-  rtc.RTCDataChannelState get lossyDataChannelState =>
-      _lossyDC?.state ?? rtc.RTCDataChannelState.RTCDataChannelClosed;
   bool _iceConnected = false;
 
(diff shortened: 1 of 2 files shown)


lib/src/core/engine.dart
```diff
@@ -192,49 +187,61 @@ class Engine extends Disposable with EventsEmittable<EngineEvent> {
...
+        await events.waitFor<EnginePublisherIceStateUpdatedEvent>(
+          filter: (event) => event.iceState.isConnected(),
+          duration: Timeouts.iceConnection,
+        );
+      }
 
-    if (publisher?.pc.iceConnectionState?.isConnected() == true) {
-      logger.warning('[$objectId] publisher is already connected');
-      return;
+      // wait for data channel to open (if not already)
```
hiroshihorie, line 224: This is where it waits for `open` state of the data channel.

boks1971: Is there some callback to indicate data channel open failure if it fails for some reason?

lib/src/core/engine.dart
```diff
@@ -192,49 +187,61 @@ class Engine extends Disposable with EventsEmittable<EngineEvent> {
...
-        packet.kind == lk_models.DataPacket_Kind.LOSSY ? _lossyDC : _reliableDC;
+    if (_subscriberPrimary) {
+      // make sure publisher transport is connected
 
-    // send if channel exists
-    if (channel == null) {
-      throw UnexpectedStateException('Data channel is not ready');
-    }
+      if (publisher?.pc.iceConnectionState?.isConnected() != true) {
+        logger.fine('Publisher is not connected...');
```
boks1971, line 209: What happens if there are a bunch of data channel messages sent by the app in this state? Are all of them becoming promises and get sent (promises resolved) when the data channel connects? Or are the messages just dropped on the floor?

lib/src/core/engine.dart
hiroshihorie: `waitFor` will throw a `TimeoutException` if it timeouts.

lib/src/core/engine.dart
hiroshihorie: promises should be pending until `waitFor` condition clears (received notification) or timeouts.
actually maybe there should be more checks to prevent `negotiate()` being called multiple times.. 🤔
(not sure how immediate the state changes to `RTCIceConnectionStateChecking`)

davidzhao (approved):
looks good!

PR #68, opened 2022-01-07
Subscription permission API

hiroshihorie:
Still needs more polishing but would appreciate if oniichans could take a look 😁

TODO
- JS is emitting `TrackEvent.Ended`
- [x] Android's `ParticipantTrackPermission` has `allTracksAllowed`
- [ ] Emit `TrackSubscribed` / `TrackUnubscribed` when permission resumed.

**Needs testing**

### DIFF
--- a/lib/src/events.dart
+++ b/lib/src/events.dart
@@ -244,3 +244,18 @@ class SpeakingChangedEvent with ParticipantEvent {
     required this.speaking,
   });
 }
+
+/// One of subscribed tracks have changed its permissions for the current
+/// participant. If permission was revoked, then the track will no longer
+/// be subscribed. If permission was granted, a TrackSubscribed event will
+/// be emitted.
+class TrackSubscriptionPermissionChangedEvent with RoomEvent, ParticipantEvent {
+  final Participant participant;
+  final RemoteTrackPublication trackPublication;
+  final TrackSubscriptionState state;
+  const TrackSubscriptionPermissionChangedEvent({
+    required this.participant,
+    required this.trackPublication,
+    required this.state,
+  });
+}
--- a/lib/src/participant/local.dart
+++ b/lib/src/participant/local.dart
@@ -312,14 +312,15 @@ class LocalParticipant extends Participant<LocalTrackPublication> {
   /// who is able to subscribe at a participant and track level.
   ///
   /// Note: if access is given at a track-level (i.e. both [allParticipantsAllowed] and
-  /// [allTracksAllowed] are false), any newer published tracks
+  /// [ParticipantTrackPermission.allTracksAllowed] are false), any newer published tracks
   /// will not grant permissions to any participants and will require a subsequent
   /// permissions update to allow subscription.
   ///
-  /// @param allParticipantsAllowed Allows all participants to subscribe all tracks.
-  /// Takes precedence over [[participantTrackPermissions]] if set to true.
+  /// [allParticipantsAllowed] Allows all participants to subscribe all tracks.
+  /// Takes precedence over [trackPermissions] if set to true.
   /// By default this is set to true.
-  /// @param participantTrackPermissions Full list of individual permissions per
+  ///
+  /// [trackPermissions] Full list of individual permissions per
   /// participant/track. Any omitted participants will not receive any permissions.
 
   void setTrackSubscriptionPermissions({
--- a/lib/src/participant/local.dart
+++ b/lib/src/participant/local.dart
@@ -3,6 +3,7 @@ import 'package:flutter_webrtc/flutter_webrtc.dart' as rtc;
 import 'package:meta/meta.dart';
 
 import '../core/room.dart';
+import '../core/signal_client.dart';
 import '../events.dart';
 import '../exceptions.dart';
 import '../extensions.dart';
@@ -304,4 +305,29 @@ class LocalParticipant extends Participant<LocalTrackPublication> {
       }
     }
   }
+
+  /// Control who can subscribe to LocalParticipant's published tracks.
+  ///
+  /// By default, all participants can subscribe. This allows fine-grained control over
+  /// who is able to subscribe at a participant and track level.
+  ///
+  /// Note: if access is given at a track-level (i.e. both [allParticipantsAllowed] and
+  /// [allTracksAllowed] are false), any newer published tracks
+  /// will not grant permissions to any participants and will require a subsequent
+  /// permissions update to allow subscription.
+  ///
+  /// @param allParticipantsAllowed Allows all participants to subscribe all tracks.
+  /// Takes precedence over [[participantTrackPermissions]] if set to true.
+  /// By default this is set to true.
+  /// @param participantTrackPermissions Full list of individual permissions per
+  /// participant/track. Any omitted participants will not receive any permissions.
+
+  void setTrackSubscriptionPermissions({
+    required bool allParticipantsAllowed,
+    List<ParticipantTrackPermission> trackPermissions = const [],
+  }) =>
+      room.engine.signalClient.sendUpdateSubscriptionPermissions(
+        allParticipants: allParticipantsAllowed,
+        trackPermissions: trackPermissions.map((e) => e.toPBType()).toList(),
+      );
 }
--- a/lib/src/publication/remote.dart
+++ b/lib/src/publication/remote.dart
@@ -279,7 +279,7 @@ class RemoteTrackPublication<T extends RemoteTrack>
   }
 
   TrackSubscriptionState subscriptionState() {
-    if (!subscribed || !super.subscribed) {
+    if (!super.subscribed) {
       return TrackSubscriptionState.unsubscribed;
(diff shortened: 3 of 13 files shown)


lib/src/types.dart
```diff
@@ -159,3 +165,19 @@ class VideoDimensions {
...
+
+@immutable
+class ParticipantTrackPermission {
+  /// The participant id this permission applies to.
+  final String participantSid;
+
+  // The list of track ids that the target participant can subscribe to.
+  // When unset, it'll allow all tracks to be subscribed by the participant.
+  // When empty, this participant is disallowed from subscribing to any tracks.
+  final List<String>? allowedTrackSids;
```
davidliu, line 177: Hmm, should probably just keep the `allTracksAllowed` allowed flag rather than this unset vs empty behavior difference. It'll keep the behavior similar across platforms.

lib/src/participant/local.dart
```diff
@@ -304,4 +305,29 @@ class LocalParticipant extends Participant<LocalTrackPublication> {
...
+  ///
+  /// By default, all participants can subscribe. This allows fine-grained control over
+  /// who is able to subscribe at a participant and track level.
+  ///
+  /// Note: if access is given at a track-level (i.e. both [allParticipantsAllowed] and
+  /// [allTracksAllowed] are false), any newer published tracks
+  /// will not grant permissions to any participants and will require a subsequent
+  /// permissions update to allow subscription.
+  ///
+  /// @param allParticipantsAllowed Allows all participants to subscribe all tracks.
```
davidliu, line 319: does Dart use the same `@param` syntax in comments as Java?

davidliu (approved):
LGTM!

lib/src/events.dart
```diff
@@ -244,3 +244,18 @@ class SpeakingChangedEvent with ParticipantEvent {
     required this.speaking,
   });
 }
+
+/// One of subscribed tracks have changed its permissions for the current
+/// participant. If permission was revoked, then the track will no longer
+/// be subscribed. If permission was granted, a TrackSubscribed event will
```
boks1971, line 250: Actually, track is still subscribed from server's point of view. When permission is revoked, server will keep the subscription in a pending list and resume the track when the permissions are granted (if the permissions are granted)

boks1971: Are you emitting `TrackSubscribed` on resume?

boks1971 (approved):

lib/src/participant/local.dart
hiroshihorie: doesn't look like it's recommended  😅 https://dart.dev/guides/language/effective-dart/documentation#do-use-prose-to-explain-parameters-return-values-and-exceptions

lib/src/types.dart
hiroshihorie: This behavior was copied from the JS SDK PR. I agree it should be similar across platforms.
https://github.com/livekit/client-sdk-js/pull/109/files#diff-0cccdd41433a8b9763e61ed69f5e048472997700549df11e3b6c6af08d15af1aR9-R14

Maybe I will follow Android's PR 🤔 Would like to hear @davidzhao 's opinion.

hiroshihorie: changed it!

lib/src/events.dart
hiroshihorie: Interesting point! I copied this comment from dz's PR https://github.com/livekit/client-sdk-js/pull/109
I might be misunderstanding something, so technically it will remain subscribed but it is desired to emit TrackSubscribed / TrackUnsubscribed ?

hiroshihorie:
It might be best if I actually test the behavior 😅

lib/src/events.dart
hiroshihorie: It might be best if I actually test the behavior 😅

lib/src/types.dart
davidzhao: I think DL makes a fair point about consistency. With JS I mainly removed it because it was quite verbose setting that field when testing. Since JS uses an interface, the user has to explicitly type out all these fields.

With that said, I agree consistency + clarity outweighs verbosity, and will change it in JS as well.

lib/src/events.dart
hiroshihorie: also I was wondering if permissions can be set at publish time ? or it's only possible after publish.

boks1971: I am almost useless here @hiroshihorie . I do not fully remember what events get emitted and the exact meaning of those.

Functionally, subscription/unsubscription action is something which the app tells the SDK explicitly. The permissions is something that is coming from the publisher side. So, because of permissions, a track pausing/resuming does not constitute unsubscription as the app did not ask for it. So, the track is subscribed and server will send media when there are permissions. If there are no permissions, it is notified via `SubscriptionPermissionUpdate`.

What I am trying to say is the notion of `subscribed` is app asking our SDK to subscribe to a track. We get that message and send media if there is permission.

I am stressing on that because apps need not do anything different in the subscription path. They just subscribe and be done with it. Until they want to unsubscribe. And the media may pause/resume because of permissions and we emit the SusbcriptionPermissionUpdate to let them know. `TrackSubscribed`/`TrackUnsubscribed` events seem like it is in response to their API and I do not think we should emit them on track pause/resume because of permissions.

boks1971: Permissions is a separate message now. It needs to be because it can change in the middle.

But, good question about SDKs providing that option at creation time and SDKs converting that into a second protocol message. WDYT @davidzhao @davidliu

davidzhao: I think it may be useful, but IMO adds cognitive load to `publishTrack`. I say let's wait and see how people are using it first before linking the two things together. Even with the current APIs, it's possible to achieve minimal permissions so there isn't "leakage" from new tracks being subscribed by unintended participants.

davidzhao: > Interesting point! I copied this comment from dz's PR [livekit/client-sdk-js#109](https://github.com/livekit/client-sdk-js/pull/109) I might be misunderstanding something, so technically it will remain subscribed but it is desired to emit TrackSubscribed / TrackUnsubscribed ?

To the client's perspective, the track will not be available for playback, and an `Unsubscribed` event will be triggered. When permissions are granted again, a `Subscribed` event should be fired again. CC: @davidliu  as well to make sure we are aligned on behavior.

Without the explicit Unsubscribed event, the client's UI may be stuck in a weird state where they are still bound to the remote track, but will not be getting any frames.

lib/src/publication/remote.dart
```diff
@@ -245,4 +249,21 @@ class RemoteTrackPublication<T extends RemoteTrack>
...
+  @internal
+  // Update internal var and return true if changed
+  bool updateSubscriptionAllowed(bool allowed) {
+    if (_subscriptionAllowed == allowed) return false;
+    _subscriptionAllowed = allowed;
+    return true;
+  }
+
+  TrackSubscriptionState subscriptionState() {
+    if (!subscribed || !super.subscribed) {
```
davidzhao, line 262: what is the difference between `subscribed` and `super.subscribed` ?

davidzhao (approved):
lg!

lib/src/publication/remote.dart
davidliu: super.subscribed is `track != null`

davidzhao: I'm not seeing a getter definition for `RemoteTrackPublication.subscribed`, only a setter. Wouldn't the two behave the same?

davidliu: TrackPublication.subscribed would be the appropriate place to look.

davidliu: Speaking of `subscribed`, @hiroshihorie, the subscribed value should take into account `subscriptionAllowed` and return false if allowed is false.

hiroshihorie:
Always was curious which is correct english `subscriptionStatus` or `subscriptionState` 😅

hiroshihorie:
I updated example UI and I'm experimenting with it now.
@boks1971 When the track is `unsubscribed`, no more `SubscriptionPermissionUpdateEvent`s seem to arrive over the signal even the publisher is changing the permissions. Is this by design ?
Only when I send `UpdateSubscription` again I can get the signal that I am `not allowed` to subscribe.

boks1971:
> I updated example UI and I'm experimenting with it now. @boks1971 When the track is `unsubscribed`, no more `SubscriptionPermissionUpdateEvent`s seem to arrive over the signal even the publisher is changing the permissions. Is this by design ? Only when I send `UpdateSubscription` (subscribe: true) again I can get the signal that I am `not allowed` to subscribe.

That sounds like a bug on server side @hiroshihorie . This is not well tested on server side. Very possible that there is a bug. I think I will have to run this to debug. Is there a way I can run this?

davidzhao:
> Always was curious which is correct english `subscriptionStatus` or `subscriptionState` 😅

Both are technically correct IMHO. I just picked one and went with it.

PR #70, opened 2022-01-10
internet permission for example's Manifest

alex-bluetrain:
if you build an APK using the example, it won't connect to the server, it's missing the INTERNET permission on AndroidManifest.xml


hiroshihorie (approved):
Nice 🙂

PR #74, opened 2022-01-26
Fix crash on android in portrait

davidliu:
encoding scale calculation was using the wrong side, and ended up getting the wrong numbers (e.g. odd numbers which crash android).


hiroshihorie (approved):
Nice fix!

PR #82, opened 2022-02-14
Ensure even video dimensions

davidliu:
Looks like there's a separate issue where the dimensions we're using for scale calculation may not match the capture format the camera is using, so the even dimensions checking we're using here may still be incorrect.

### DIFF
--- a/lib/src/types.dart
+++ b/lib/src/types.dart
@@ -159,6 +159,9 @@ class VideoDimensions {
   /// Returns the larger value
   int max() => math.max(width, height);
 
+  /// Returns the smaller value
+  int min() => math.min(width, height);
+
   VideoDimensions copyWith({
     int? width,
     int? height,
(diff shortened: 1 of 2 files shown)


hiroshihorie (approved):

davidliu:
@hiroshihorie do you know if Swift has a limited set of capture formats as well? I think we may need to address that there as well.

hiroshihorie:
Do you mean the dimensions provided by the camera ? Yes there are only a few to choose from.
``["CMVideoDimensions(width: 640, height: 480)", "CMVideoDimensions(width: 1280, height: 720)", "CMVideoDimensions(width: 1920, height: 1080)", "CMVideoDimensions(width: 1080, height: 1920)", "CMVideoDimensions(width: 1760, height: 1328)", "CMVideoDimensions(width: 1328, height: 1760)", "CMVideoDimensions(width: 1552, height: 1552)"]``

The SDK choses the smallest that satisfies the required dimensions.

hiroshihorie:
Thats really great to know m97 does a better job 🙂, in my opinion I don't think we need to do adjusting etc in that case.
But I would like to check with both h264 and vp8 since they behave very differently.

PR #83, opened 2022-02-15
Revert "Ensure even video dimensions"

davidliu:
Reverts livekit/client-sdk-flutter#82


davidliu:
OMG, I accidentally hit squash and merge instead of the comment I wrote to the previous PR.

It looks like m97 handles the odd dimensions now, so I think we won't need this fix.

hiroshihorie:
I agree if everything works fine, we might not need this 🙂

hiroshihorie (approved):

PR #85, opened 2022-02-21
Update presets

hiroshihorie:


lukasIO (approved):

hiroshihorie:
Take your time @lukasIO 😂

lib/src/track/options.dart
```diff
@@ -113,6 +113,194 @@ extension VideoEncodingExt on VideoEncoding {
...
+  static const presetH2160_169 = VideoParameters(
+    dimensions: VideoDimensions.h2160_169,
+    encoding: VideoEncoding(
+      maxBitrate: 8 * 1000 * 1000,
+      maxFramerate: 30,
+    ),
+  );
+
+  static const presetH720_169 = VideoParameters(
+    dimensions: VideoDimensions.h540_169,
```
davidzhao, line 181: Typo? I think this should be h720.. also move to below h540? :)

hiroshihorie: 😅🙏🏻

PR #87, opened 2022-02-25
Update param computation

hiroshihorie:


lib/src/utils.dart
```diff
@@ -233,11 +280,12 @@ class Utils {
       if (i >= videoRids.length) {
         return;
       }
+      final size = dimensions.min();
       final rid = videoRids[i];
 
       result.add(e.encoding.toRTCRtpEncoding(
         rid: rid,
-        scaleResolutionDownBy: findEvenScaleDownBy(dimensions, e.dimensions),
+        scaleResolutionDownBy: size / e.dimensions.min(),
```
lukasIO, line 288: Only realizing this now, but maybe we should enforce a minimum here, so that `scaleResolutionDownBy` never gets smaller than 1.0 ?

lukasIO (approved):

lib/src/utils.dart
hiroshihorie: sure!

hiroshihorie: @davidliu I have removed findEvenScaleDownBy for now since it looks ok with M97, what do you think ?

davidliu: Yeah, that should be fine.

PR #89, opened 2022-03-03
add ability to select capture device ids

davidliu:


hiroshihorie (approved):
✅ Nice, except it won't probably work with iOS 👉 https://github.com/flutter-webrtc/flutter-webrtc/blob/8a590fe860685de4445faa0dc237b060577b9e33/common/darwin/Classes/FlutterRTCMediaStream.m#L257
for iOS I don't know why it's `sourceId`...

davidliu:
I think it should work; all the other clients are actually still using sourceId (maybe it's from the old WebRTC apis which `flutter-webrtc` still conforms to). deviceId is copied to `constraints.deviceId` (current WebRTC api) and `constraints.optional.sourceId`.

hiroshihorie:
ah didn't know it gets copied, nice!

lukasIO (approved):

PR #100, opened 2022-04-02
Simplify internal events

hiroshihorie:
- Remove redundant internal event
- `Room` also directly listens to `SignalEvents` instead of `Engine` relaying them.


davidzhao (approved):
looks good!

PR #101, opened 2022-04-03
[Protocol v7] Remote unpublish & expose permissions

hiroshihorie:


lib/src/participant/local.dart
```diff
@@ -339,4 +339,18 @@ class LocalParticipant extends Participant<LocalTrackPublication> {
   @internal
   Iterable<lk_rtc.TrackPublishedResponse> publishedTracksInfo() =>
       trackPublications.values.map((e) => e.toPBTrackPublishedResponse());
+
+  @internal
+  @override
+  bool setPermissions(ParticipantPermissions newValue) {
+    final didUpdate = super.setPermissions(newValue);
```
davidzhao, line 346: In JS, we are passing back the [previous value](https://github.com/livekit/client-sdk-js/blob/main/src/room/events.ts#L233).. since you have an event.. perhaps we can pass back both?

hiroshihorie: sure, like oldValue newValue ?

davidzhao: since you have permissions, maybe just add `oldPermissions` ?

davidzhao (approved):

lib/src/participant/local.dart
hiroshihorie: got it

PR #104, opened 2022-04-07
Change AudioSession configuration logic

hiroshihorie:
With this change, mute/unmute will not switch audio category just like Swift SDK.
Un-publishing local audio track will stop the mic indicator.

### DIFF
--- a/lib/src/track/local/local.dart
+++ b/lib/src/track/local/local.dart
@@ -1,10 +1,10 @@
 import 'package:flutter/material.dart';
-
 import 'package:flutter_webrtc/flutter_webrtc.dart' as rtc;
 import 'package:meta/meta.dart';
 
 import '../../events.dart';
 import '../../exceptions.dart';
+import '../../extensions.dart';
 import '../../logger.dart';
 import '../../participant/remote.dart';
 import '../../proto/livekit_models.pb.dart' as lk_models;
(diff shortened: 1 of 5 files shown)


lib/src/track/local/local.dart
```diff
@@ -170,4 +173,32 @@ abstract class LocalTrack extends Track {
     // mark as started
     await start();
   }
+
+  @internal
+  @mustCallSuper
+  Future<bool> publish() async {
```
davidliu, line 179: Should rename to `onPublish()` just to be clear that this doesn't actually do the track publishing itself.

lib/src/track/local/local.dart
```diff
@@ -170,4 +173,32 @@ abstract class LocalTrack extends Track {
...
+
+    logger.fine('$objectId.publish()');
+
+    _published = true;
+    return true;
+  }
+
+  @internal
+  @mustCallSuper
+  Future<bool> unpublish() async {
```
davidliu, line 193: Rename to `onUnpublish()`

davidliu (approved):
LGTM, just some naming changes suggested.

lib/src/track/local/local.dart
hiroshihorie: i agree !

PR #106, opened 2022-04-11
fix(connect): await engine.connect in async context

isqad:


davidzhao:
@isqad I believe both approaches are identical. we are returning a Future<void> from `connect`. Am I missing anything?

isqad:
I though that `async` modificator is not needed if no any `await` calls 🤔

davidzhao:
> I though that `async` modificator is not needed if no any `await` calls 🤔

that's true :) but having it is also a valid syntax.

isqad:
> but having it is also a valid syntax.

Ok) So this PR might be closed)

hiroshihorie:
Thanks @isqad I understand this looked weird, in the process of making this PR https://github.com/livekit/client-sdk-flutter/pull/109/files that code has been modified to be more clean.

isqad:
All right, It is become more clear for me, thank you!

PR #123, opened 2022-06-01
Autofill URL and TOKEN in debug mode.

cloudwebrtc:
Autofill `URL` and `TOKEN` when running in debug mode using mobile devices.

```dart
flutter run --dart-define=URL=ws://192.168.1.141:7880  --dart-define=TOKEN=eyJhbGciOiJIU......
```


hiroshihorie (approved):

hiroshihorie (approved):
Nice but maybe the env can be higher priority than the stored one.

PR #124, opened 2022-06-01
Add `audioBitrate` option for publishAudioTrack.

cloudwebrtc:
will test later, before merging I will need to verify that the maxBitrate parameter is passed to the native sdk normally.

### DIFF
--- a/lib/src/options.dart
+++ b/lib/src/options.dart
@@ -105,6 +105,12 @@ class VideoPublishOptions {
       '${runtimeType}(videoEncoding: ${videoEncoding}, simulcast: ${simulcast})';
 }
 
+class AudioPreset {
+  static const telephone = 12000;
+  static const speech = 20000;
+  static const music = 32000;
+}
+
 /// Options used when publishing audio.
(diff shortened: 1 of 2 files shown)


lib/src/options.dart
```diff
@@ -105,15 +105,25 @@ class VideoPublishOptions {
...
+
 /// Options used when publishing audio.
 class AudioPublishOptions {
   /// Whether to enable DTX (Discontinuous Transmission) or not.
   /// https://en.wikipedia.org/wiki/Discontinuous_transmission
   /// Defaults to true.
   final bool dtx;
 
+  /// max audio bitrate
+  final int audioBitrate;
```
hiroshihorie, line 122: Maybe this can be simply `bitrate` ? since it's already in `AudioPublishOptions` ?

cloudwebrtc: Makes sense, will update in a while. 
maybe it can be called `maxBitrate`, which is a limit parameter.

hiroshihorie: makes sense !

PR #128, opened 2022-06-11
re-send tracks permissions after reconnected.

cloudwebrtc:


hiroshihorie (approved):
According to @lukasIO , I think we need to send it for normal connect also ?
(set permission before connection)

davidzhao:
> According to @lukasIO , I think we need to send it for normal connect also ? (set permission before connection)

Responding since Lukas is on vacation this week. He's saying that the client could buffer up permission updates, which makes it easier for clients to use. Right now the clients have to wait until the connection is complete before setting permission

PR #130, opened 2022-06-27
handle combined participant update.

cloudwebrtc:
https://github.com/livekit/client-sdk-js/pull/275

### DIFF
--- a/lib/src/core/signal_client.dart
+++ b/lib/src/core/signal_client.dart
@@ -168,13 +168,13 @@ class SignalClient extends Disposable with EventsEmittable<SignalEvent> {
             participants: msg.update.participants));
         break;
       case lk_rtc.SignalResponse_Message.trackPublished:
-        events.emit(SignalLocalTrackPublishedEvent(
+        emitWhenConnected(SignalLocalTrackPublishedEvent(
           cid: msg.trackPublished.cid,
           track: msg.trackPublished.track,
         ));
         break;
       case lk_rtc.SignalResponse_Message.trackUnpublished:
-        events.emit(SignalTrackUnpublishedEvent(
+        emitWhenConnected(SignalTrackUnpublishedEvent(
           trackSid: msg.trackUnpublished.trackSid,
         ));
         break;
@@ -186,7 +186,7 @@ class SignalClient extends Disposable with EventsEmittable<SignalEvent> {
         events.emit(SignalRoomUpdateEvent(room: msg.roomUpdate.room));
         break;
       case lk_rtc.SignalResponse_Message.connectionQuality:
-        events.emit(SignalConnectionQualityUpdateEvent(
+        emitWhenConnected(SignalConnectionQualityUpdateEvent(
           updates: msg.connectionQuality.updates,
         ));
         break;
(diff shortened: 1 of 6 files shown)


test/core/room_e2e_test.dart
```diff
@@ -36,10 +36,16 @@ void main() {
     test('participant join', () async {
       expect(
         room.events.streamCtrl.stream,
-        emits(predicate<ParticipantConnectedEvent>(
-          (event) => event.participant.sid == remoteParticipantData.sid,
-        )),
+        emitsInOrder(<Matcher>[
```
davidzhao, line 39: is this checking to see TrackPublishedEvent is emitted before ParticipantConnectedEvent? I think that's the wrong behavior.

lib/src/participant/remote.dart
```diff
@@ -167,15 +166,13 @@ class RemoteParticipant extends Participant<RemoteTrackPublication> {
...
-        );
-        [events, room.events].emit(event);
-      }
+    // always emit events for new publications, Room will not forward them unless it's ready
+    for (final pub in newPubs) {
+      final event = TrackPublishedEvent(
+        participant: this,
+        publication: pub,
+      );
+      [events, room.events].emit(event);
```
davidzhao, line 175: I don't believe we want to do this here. TrackPublished events should not be emitted before the Room is connected.

When a room is already connected
* fire `TrackPublished`

When a room is not yet connected
* do *not* fire `TrackPublished`

davidzhao (changes requested):

lib/src/participant/remote.dart
cloudwebrtc: If the room here is not connected, do we need to cache the TrackPublished event until room is reconncted and then emit it?

davidzhao: I don't think we need to cache it

davidzhao: the event is only fired for tracks that are published after the participant has joined

cloudwebrtc: so, here we need to confirm that the `subscriber.pc` is connected, then the event can be emitted, is this correct?

lib/src/core/signal_client.dart
```diff
@@ -240,6 +240,13 @@ class SignalClient extends Disposable with EventsEmittable<SignalEvent> {
       _updateConnectionState(ConnectionState.disconnected);
     }
   }
+
+  @internal
+  void emitWhenConnected(SignalEvent event) {
```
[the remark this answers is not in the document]
cloudwebrtc: okay, I'll move the code to the correct location.

lib/src/participant/remote.dart
davidzhao: not exactly.. to be clear, when I say "connected", it's referring to Room.state == connected. This is the contract we have with developers

cloudwebrtc: ok, I'm trying to figure out the timing,  there are two `_updateConnectionState(ConnectionState.connected);` in core/engine.dart that let me know the room is connected. 
so, no more questions

davidzhao (approved):

PR #134, opened 2022-06-30
protocol 8.

cloudwebrtc:
I did some tests, we can connect the room in the following two ways

```dart
      final room = Room();
      final listener = room.createListener();

```

a)
``` dart
      await room.connect(
        _url,
        _token,
        roomOptions: RoomOptions(
           ...
        ),
        fastConnectOptions: _fastConnect
            ? FastConnectOptions(
                microphone: const TrackOption(enabled: true),
                camera: const TrackOption(enabled: true),
              )
            : null,
      );

```

b)
```dart
     final localAudioTrack = await LocalAudioTrack.create();
     final localCamTrack = await LocalVideoTrack.createCameraTrack();

      await room.connect(
        _url,
        _token,
        roomOptions: RoomOptions(
           ...
        ),
        fastConnectOptions: _fastConnect
            ? FastConnectOptions(
                microphone: TrackOption(track: localAudioTrack),
                camera:  TrackOption(track: localCamTrack),
              )
            : null,
      );

```

```

### DIFF
--- a/lib/src/core/room.dart
+++ b/lib/src/core/room.dart
@@ -13,6 +13,8 @@ import '../participant/remote.dart';
 import '../proto/livekit_models.pb.dart' as lk_models;
 import '../proto/livekit_rtc.pb.dart' as lk_rtc;
 import '../support/disposable.dart';
+import '../track/local/audio.dart';
+import '../track/local/video.dart';
 import '../track/track.dart';
 import '../types/other.dart';
 import 'engine.dart';
@@ -114,8 +116,6 @@ class Room extends DisposableChangeNotifier with EventsEmittable<RoomEvent> {
   Future<void> connect(
     String url,
     String token, {
-    bool? camera,
-    bool? microphone,
     ConnectOptions? connectOptions,
     RoomOptions? roomOptions,
     FastConnectOptions? fastConnectOptions,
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: lib/src/core/room.dart [hunk 1]
```diff
@@ -141,6 +145,32 @@ class Room extends DisposableChangeNotifier with EventsEmittable<RoomEvent> {
         info: event.response.participant,
       );
 
+      if (connectOptions.protocolVersion.index >= ProtocolVersion.v8.index &&
```
davidliu, line 148: nit: probably should be moved elsewhere, since this is for listener setup.

davidliu (approved):

lib/src/livekit.dart
```diff
@@ -9,6 +9,8 @@ class LiveKitClient {
   /// Convenience method for connecting to a LiveKit server.
   /// Returns a [Room] upon a successful connect or throws when it fails.
   /// Alternatively, it is possible to instantiate [Room] and call [Room.connect] directly.
+  @Deprecated(
+      'Use `Room.connect` instead, This method is deprecated above Protocol v8.')
   static Future<Room> connect(
     String url,
     String token, {
```
davidliu, line 16: Can we to default to protocol 7 when using the deprecated connect?

cloudwebrtc: ah, let me check this.

boks1971 (approved):

lib/src/livekit.dart
cloudwebrtc: ok, solved

PR #149, opened 2022-08-15
Fixed UI stuck when get thumbnails on screen sharing.

cloudwebrtc:


davidzhao (approved):
lgtm

lib/src/widgets/screen_select_dialog.dart
```diff
@@ -42,38 +110,43 @@ class ScreenSelectDialog extends Dialog {
...
+      for (var element in sources) {
+        if (kDebugMode) {
+          print(
+              'name: ${element.name}, id: ${element.id}, type: ${element.type}');
         }
-      });
+      }
       _timer?.cancel();
-      _timer = Timer.periodic(const Duration(seconds: 2), (timer) {
+      _timer = Timer.periodic(const Duration(seconds: 3), (timer) {
```
davidzhao, line 137: one concern is with this running on slower computers, if it takes more than 3s to update complete a set of thumbnails, it could just be doing this constantly?

cloudwebrtc: Here timer actually only refresh the source list, add or remove, the thumbnail is done in another thread, only the content of the window is changed (image changed), and the thumbnail change event will be emitted.
https://github.com/webrtc-sdk/webrtc/pull/37/files#diff-2ed6dd2f4d3d1562a1c0810e03c7ae6f504a4472009708ed62a538d2997893ebR157

davidzhao: ah ok, understood. thank you for the explanation.

PR #155, opened 2022-08-22
fix: close issue #154.

cloudwebrtc:
fix: The local participant widget cannot respond to the muted state


cloudwebrtc:
This fixes the local mic, the camera cannot be displayed correctly, so merge it first

PR #156, opened 2022-08-24
feat: network type detection.

cloudwebrtc:


lib/src/utils.dart
```diff
@@ -286,6 +289,33 @@ class Utils {
...
+      case ConnectivityResult.bluetooth:
+        networkType = 'bluetooth';
+        break;
+      case ConnectivityResult.ethernet:
+        networkType = 'ethernet';
+        break;
+      case ConnectivityResult.none:
+        networkType = 'none';
+        break;
+    }
```
hiroshihorie, line 315: I think we should follow the spec in the proto?
// wifi, wired, cellular, vpn, empty if not known

cloudwebrtc: sure,  I can change it to be consistent
mobile -> cellular
ethernet -> wired
none -> empty

But it seems https://pub.dev/packages/connectivity_plus plugin can't detect VPN, and has a `bluetooth` type

hiroshihorie: yes I can't detect VPN also... I'm just sending empty for now.

hiroshihorie (approved):

PR #157, opened 2022-08-26
Add ping/pong.

cloudwebrtc:


lib/src/core/signal_client.dart
```diff
@@ -240,6 +256,47 @@ class SignalClient extends Disposable with EventsEmittable<SignalEvent> {
       _updateConnectionState(ConnectionState.disconnected);
     }
   }
+
+  void _sendPing() {
+    _sendRequest(lk_rtc.SignalRequest()
```
davidzhao, line 261: I think we need to call `_resetPingTimeout` here; each time we send a ping. otherwise subsequent ping timeouts would not be detected.

cloudwebrtc: that makes sense

lib/src/core/signal_client.dart
```diff
@@ -223,6 +236,9 @@ class SignalClient extends Disposable with EventsEmittable<SignalEvent> {
       case lk_rtc.SignalResponse_Message.notSet:
         logger.info('signal message not set');
         break;
+      case lk_rtc.SignalResponse_Message.pong:
+        _resetPingTimeout();
```
davidzhao, line 240: should it be `_clearPingTimeout` here? that means ping had been successful. without setting a new one.

davidzhao: oops sorry.. I think I'm confusing myself.

cloudwebrtc: I think either doing _clearPingTimeout or _resetPingTimeout will work

If the pingTimer is cleared, a new one will be created when sendPing
If we keep _resetPingTimeout, we don't need to call _resetPingTimeout again when we sendPing

Issue #163, opened 2022-09-07
Web - AudioCaptureOptions cant use deviceId and optional parameters


jdde:
When you're initiating a track in the flutter web version and provide AudioCaptureOptions, the MediaConstraintsMap is generated in a way, which is not compatible with current web implementations.

To reproduce the problem, just try to getUserMedia in your current browser (I'm using chrome v104)
```
window.navigator.mediaDevices.getUserMedia({
    audio: {
      optional: [{echoCancellation: true}],
      deviceId: 'default'
    }
})
```
This will throw an error:
TypeError: Failed to execute 'getUserMedia' on 'MediaDevices': Malformed constraint: Cannot use both optional/mandatory and specific or advanced constraints.

The same error is thrown, when you initialize an audio track with the following AudioCaptureOptions:
```
AudioCaptureOptions(
  noiseSuppression: audioDefaultNoiseSuppression,
  autoGainControl: audioDefaultAutoGainControl,
  highPassFilter: audioDefaultHighPassFilter,
  typingNoiseDetection: audioDefaultTypingNoiseDetection,
  deviceId: 'default'
)
```
That's because the toMediaConstraintsMap() function add optional parameters AND the deviceId.
When you just pass one of them, the web api accepts it.

My suggestion would be to not set optional parameters, when you're providing a mandatory parameter in web.
Since those options are not supported anyway.

The same problem exists for video tracks.

jdde:
update: checked the video implementation. It's not providing any optional parameters, so it will work.

jdde:
https://github.com/livekit/client-sdk-flutter/pull/164

jdde:
pull request was merged :)

PR #164, opened 2022-09-07
Web - AudioCaptureOptions cant use deviceId and optional parameters

jdde:
https://github.com/livekit/client-sdk-flutter/issues/163

- do not set mandatory & optional parameters, when used in web
- web is not allowing it
- adjusted logic to use deviceId over optional parameters in web


cloudwebrtc (approved):
lgtm

PR #174, opened 2022-09-15
Make MediaDevice data class

jascodes:
It's useful for Media device to be data class for example in case of selection box.


jascodes:
@davidliu Any comments?

cloudwebrtc (approved):
lgtm

FreddyKaiser:
Any reason why [groupid](https://developer.mozilla.org/en-US/docs/Web/API/MediaDeviceInfo/groupId ) has not been taken in consideration ?

PR #177, opened 2022-09-26
Supports getting the connected remote address.

cloudwebrtc:
Supports getting the connected remote address for verification migration.

Tested platform.
- [x] Android
- [x] macOS
- [x] Web
- [x] Windows
- [x] iOS

### DIFF
--- a/lib/src/core/engine.dart
+++ b/lib/src/core/engine.dart
@@ -758,10 +758,10 @@ Future<String?> getConnectedAddress(rtc.RTCPeerConnection pc) async {
       case 'candidate-pair':
       case 'googCandidatePair':
         if (selectedCandidatePairId == '') {
-          if ((v.values['selected'] != null && v.values['selected'] == true)) {
+          if (v.values['selected'] != null && v.values['selected'] == true) {
             selectedCandidatePairId = v.id;
-          } else if ((v.values['googActiveConnection'] != null &&
-              v.values['googActiveConnection'] == true)) {
+          } else if (v.values['googActiveConnection'] != null &&
+              v.values['googActiveConnection'] == true) {
             selectedCandidatePairId = v.id;
           }
         }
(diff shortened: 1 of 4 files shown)


cloudwebrtc:
getting `:7882` on iOS doesn't seem to get an IP address, is it due to some security mechanism, or is it a bug of the WebRTC SDK on iOS?

image

lib/src/core/engine.dart
```diff
@@ -709,3 +742,59 @@ extension EngineInternalMethods on Engine {
...
+  // id -> candidate ip
+  final candidates = <String, String>{};
+  final List<rtc.StatsReport> stats = await pc.getStats();
+  for (var v in stats) {
+    switch (v.type) {
+      case 'transport':
+      case 'googComponent':
+        selectedCandidatePairId = v.values['selectedCandidatePairId'] as String;
+        break;
+      case 'candidate-pair':
```
davidzhao, line 758: how come they are still coming through via these different variations? Isn't the same libwebrtc that it's building against?

davidzhao:
> getting `:7882` on iOS doesn't seem to get an IP address, is it due to some security mechanism, or is it a bug of the WebRTC SDK on iOS?

my guess it's a bug. IMO fine to skip the IP check for the purpose of the test.

lib/src/core/engine.dart
cloudwebrtc: I haven't migrated to the standard API of getStats yet, because the Android SDK misses getStats for RtpSender/RtpReceiver, and needs a small patch for libwebrtc Android.

davidzhao: I see, seems a bit messy to propagate this to clients of the library, but good to know how to clean this up in the future.

davidzhao (approved):

Issue #184, opened 2022-10-05
After refreshing LiveKit server to 1.2.3  SDK can't connect to server


janoskranczler:
After refreshing LiveKit server to 1.2.3  SDK can't connect to server. Neither Android or iOS.

Flutter version:
[✓] Flutter (Channel stable, 3.3.3, on macOS 12.6 21G115 darwin-arm, locale en-HU)
    • Flutter version 3.3.3 on channel stable at /Users/janoskranczler/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 18a827f393 (7 days ago), 2022-09-28 10:03:14 -0700
    • Engine revision 5c984c26eb
    • Dart version 2.18.2
    • DevTools version 2.15.0

Plugin version: 1.1.4

davidzhao:
How is your server configured? Is it using a single port or a wider port range?

janoskranczler:
Using a wider port range: 50000-60000.

cloudwebrtc:
After testing for half an hour, there was an occasional failure to connect. It seems to be a deadlock of server signaling. The client must be killed and restarted before the client can connect.

cloudwebrtc:
but I found another issue when I turned on Ethernet and WIFI on mac at the same time, the Android client connection timed out (the ICE connection could not be established, maybe it is related to the default route on the mac),

janoskranczler:
Thank you for your help in advance. We are using Redis. We can't connect with a client when we update the server to 1.2.3 (from 1.2.2).
```2022-10-12T13:06:35.097Z	INFO	livekit	rtc/participant_signal.go:181	could not send message to participant	{"room": "1pnjef2fic4r6p0m", "roomID": "RM_W2iS8TVzCFm8", "participant": "8smj8i2w44cf2h53", "pID": "PA_FTJt286V4KQg", "remote": false, "message": "*livekit.SignalResponse_Leave", "error": "no response sink"}```
Our RTC configuration:
```  rtc:
    allow_tcp_fallback: true
    use_ice_lite: false
    use_external_ip: false
    port_range_start: 50000
    port_range_end: 60000
    tcp_port: 7801```

cloudwebrtc:
I looked at this error and it seems to be related to the `response_sink` not being found, can you try turning off redis and testing again?

https://github.com/livekit/livekit/blob/v1.2.3/pkg/rtc/participant_signal.go#L181
[...]

PR #192, opened 2022-10-18
fix ice config issues.

cloudwebrtc:
close https://github.com/livekit/client-sdk-flutter/issues/184


lib/src/extensions.dart
```diff
@@ -24,7 +24,7 @@ extension ICEServerExt on lk_rtc.ICEServer {
   RTCIceServer toSDKType() => RTCIceServer(
         urls: urls,
         username: username.isNotEmpty ? username : null,
-        credential: credential.isNotEmpty ? username : null,
+        credential: credential.isNotEmpty ? credential : null,
```
davidzhao, line 27: oh wow.. great catch!! this would explain why users had been reporting TURN issues when running flutter.

davidzhao (approved):

PR #194, opened 2022-10-24
e2e: fix resume/full-reconnect.

cloudwebrtc:


davidliu:
What's the source issue this commit is fixing?

cloudwebrtc:
Fixed full-reconnect failure when simulating migration/NodeFailure,
For example, the server sendLeave carries the canReconnect=true flag to the client, or the signaling error or the `peerConnection` fails. At this time, the client needs to `restartConnection`, recreate pub/sub peerconnection, and republish/subscribe tracks.

davidzhao:
adding @lukasIO to take a look as this seems like a more direct port of the same logic from JS SDK.

cloudwebrtc:
Yes, I want to follow the logic of js (because js has passed the e2e test), I should be able to re-run e2e-flutter at night.

lukasIO:
From my perspective (without having actually run that code) it looks good! 
One thing that we recently added to the js reconnect logic is an event listener to detect when the browser goes online again. 
The pitifall with the `ReconnectPolicy` being timer based is that it will get throttled intensively (esp. in chrome) when backgrounded/sleeping etc. Not sure how other platforms will handle this. 
The workaround on the js side was to retry connection immediately when the browser detects that the online status has changed instead of waiting for the timer to trigger.

x3y5:
i am looking forward to this function, still waiting for the review :) @cloudwebrtc @davidzhao @lukasIO @hiroshihorie

cloudwebrtc:
I think it's time to review now, last week I passed the tests for desktop/web, and iOS real devices, android will have a crash caused by EGL base context after creating too many renderers on real devices, but I think it should be another issue because all e2ee tests pass on the android emulator.

lib/src/core/engine.dart
```diff
@@ -145,14 +148,14 @@ class Engine extends Disposable with EventsEmittable<EngineEvent> {
...
-        onTimeout: () => throw ConnectException(),
+        onTimeout: () => throw ConnectException(
+            'Timed out waiting for EnginePeerStateUpdatedEvent'),
       );
 
       _updateConnectionState(ConnectionState.connected);
     } catch (error) {
       logger.fine('Connect Error $error');
       _updateConnectionState(ConnectionState.disconnected);
-      rethrow;
```
davidzhao, line 155: does this mean errors are no longer reported to the caller? how does Room know that the connection had errored?

lib/src/core/engine.dart
```diff
@@ -433,9 +373,19 @@ class Engine extends Disposable with EventsEmittable<EngineEvent> {
       final isPrimaryOrPublisher = event.isPrimary ||
           (_hasPublished && event is EnginePublisherPeerStateUpdatedEvent);
 
-      if (isPrimaryOrPublisher && event.state.isDisconnectedOrFailed()) {
+      if (isPrimaryOrPublisher &&
```
davidzhao, line 376: I'm not sure if `isPrimaryOrPublisher` is the right logic. if either peer connection is disconnected, I think we should trigger a handleDisconnect sequence.

the handling could be much simpler.

lib/src/core/engine.dart
```diff
@@ -433,9 +373,19 @@ class Engine extends Disposable with EventsEmittable<EngineEvent> {
...
+      }
+
+      if (isPrimaryOrPublisher &&
+          event.state.isDisconnectedOrFailed() &&
+          pcState ==
+              rtc.RTCPeerConnectionState.RTCPeerConnectionStateConnected) {
         // trigger reconnect sequence
-        _onDisconnected(DisconnectReason.peerConnection);
+        pcState = event.state;
+        handleDisconnect(DisconnectReason.peerConnectionClosed);
```
davidzhao, line 388: this seems to confuse close vs disconnected/failed. in JS, we are only triggering a full reconnect when the peer connection had gone to `closed`, but here it seems to be treating a disconnected or failed event as closed.

lib/src/core/engine.dart
```diff
@@ -599,21 +549,137 @@ class Engine extends Disposable with EventsEmittable<EngineEvent> {
...
+      bool reConnectSignal = [
+        DisconnectReason.leaveReconnect,
+        DisconnectReason.peerConnectionClosed,
+        DisconnectReason.signal
+      ].contains(reason);
+      await resumeConnection(reConnectSignal);
+    }
+  }
+
+  Future<void> resumeConnection([bool reConnectSignal = false]) async {
```
davidzhao, line 589: this seems incorrect as well. in JS, even with a resume, it should *always* reconnect to signal. The difference is:
* `reconnect: false` when it's a full reconnect
* `reconnect: true` when it's a resume (ICE restart or migration)

we always reconnect signal in case the old server has failed.

davidzhao: in JS, we take in a parameter in order to know whether or not to emit a `Resuming` event.

lib/src/core/transport.dart
```diff
@@ -126,10 +127,18 @@ class Transport extends Disposable {
       }
     }
 
+    if (restartingIce && !rtc.WebRTC.platformIsWeb) {
```
davidzhao, line 130: how do we handle this for web?

lib/src/participant/local.dart
```diff
@@ -240,6 +243,21 @@ class LocalParticipant extends Participant<LocalTrackPublication> {
     await pub.dispose();
   }
 
+  Future<void> rePublishAllTracks({bool notify = true}) async {
+    final tracks = trackPublications.values.toList();
+    for (LocalTrackPublication track in tracks) {
+      await unpublishTrack(track.sid, notify: notify, stopOnUnpublish: false);
```
davidzhao, line 249: this might not be needed, because the old server/peer connection is gone.

we do, however, need to republish it to the new room.

lib/src/participant/remote.dart
```diff
@@ -190,7 +190,8 @@ class RemoteParticipant extends Participant<RemoteTrackPublication> {
   }
 
   @override
-  Future<void> unpublishTrack(String trackSid, {bool notify = true}) async {
+  Future<void> unpublishTrack(String trackSid,
+      {bool notify = true, bool? stopOnUnpublish}) async {
```
davidzhao, line 194: `stopOnUnpublish` is unused. we need to avoid calling `track.stop()` when this is true.

lib/src/core/transport.dart
cloudwebrtc: On the web, it will be triggered by `pc.createOffer({iceRestart: true});`, only in flutter native need to call `pc.restartIce();`

lib/src/core/engine.dart
cloudwebrtc: My intention to add reConnectSignal here is to trigger an iceRestart to reconnect the peerconnection when the signaling is connected but the ice connection is temporarily disconnected.

davidzhao: we are not able to distinguish peerconnection temporarily disconnected vs the server is gone. Because of that, we should always reconnect the signal connection because it will start performing a media server discovery.

cloudwebrtc: okay, makes sense

lib/src/participant/remote.dart
cloudwebrtc: In fact, I am a little confused here, unpublishTrack actually operates RemoteTrack, and the function name sounds like it should control the publishing track.

cloudwebrtc: seems to be changed to 
`Future<void> unSubscribeRemoteTrack(String trackSid, {bool notify = true, bool? stopOnUnSubscribe});` 
more reasonable?

davidzhao: oh you are right, the naming is quite confusing here. it should be something like `removePublishedTrack(String trackSid, {bool notify = true})`

davidzhao: or maybe `removePublication`

lib/src/core/engine.dart
cloudwebrtc: resolved https://github.com/livekit/client-sdk-flutter/pull/194/commits/678c430949d559ce6b1bcbda525a6a74fedcbe13

lib/src/participant/remote.dart
cloudwebrtc: https://github.com/livekit/client-sdk-flutter/pull/194/commits/18f94d03ace109964b84702f276814e26b5922fd

lib/src/core/engine.dart
cloudwebrtc: https://github.com/livekit/client-sdk-flutter/pull/194/commits/17396100604570b50a833c0f54fe333a1b040dbd

lib/src/core/engine.dart
cloudwebrtc: resolved https://github.com/livekit/client-sdk-flutter/pull/194/commits/430bfa067cd6f8644973c2aa6701dadb01c08fe3

davidzhao (approved):
looks good! nice work cleaning this up.

PR #226, opened 2022-12-23
e2ee.

cloudwebrtc:
verified platforms
- [x] macOS
- [x] iOS
- [x] Android
- [x] Web
- [x] Windows 

supported codec
- [x] VP8
- [x] H264

other features
- [x] Simulcast with E2EE
- [x] Flutter example app e2ee option
- [x] Transceiver re-use during subscription (native)
- [x] Support string key
- [x] Fire events to indicate decryption failure
- [ ] Multi-codec simulcast, might need another PR？


lib/src/e2ee/options.dart
```diff
@@ -0,0 +1,13 @@
+import 'key_provider.dart';
+
+enum EncryptionType {
+  kNone,
+  kGcm,
+  kCutsom,
```
theomonnom, line 6:
```suggestion
  kCustom,
```
small typo here

lib/src/e2ee/key_provider.dart
```diff
@@ -0,0 +1,90 @@
+import 'dart:typed_data';
+
+import 'package:flutter_webrtc/flutter_webrtc.dart';
+import 'package:cryptography/cryptography.dart';
+
+const defaultRatchetSalt = 'LKFrameEncryptionKey';
```
theomonnom, line 6: Seems like there's no way to change the salt? (yet)

theomonnom (approved):
lgtm!

lib/src/e2ee/e2ee_manager.dart
```diff
@@ -0,0 +1,150 @@
...
+  Room? _room;
+  final Map<String, FrameCryptor> _frameCryptors = {};
+  final BaseKeyProvider _keyProvider;
+  final Algorithm _algorithm = Algorithm.kAesGcm;
+  bool _enabled = true;
+  EventsListener<RoomEvent>? _listener;
+  E2EEManager(this._keyProvider);
+
+  Future<void> setup(Room room) async {
+    if (_room != room) {
```
davidzhao, line 21: does it need to deregister from the previous room?

lib/src/e2ee/events.dart
```diff
@@ -0,0 +1,30 @@
...
+    required this.publication,
+    required this.state,
+  });
+
+  @override
+  String toString() => '${runtimeType}'
+      '(participant: ${participant}, publication: ${publication}, state: ${state})';
+
+  @Deprecated('Use publication instead')
+  TrackPublication get track => publication;
```
davidzhao, line 29: since this is a new event, we can remove the deprecated getter.

lib/src/e2ee/utils.dart
```diff
@@ -0,0 +1,5 @@
+import 'package:flutter_webrtc/flutter_webrtc.dart';
+
+bool isE2EESupported() {
+  return WebRTC.platformIsMobile || WebRTC.platformIsDesktop;
```
davidzhao, line 4: would this be accurate for web?

lib/src/participant/local.dart
```diff
@@ -271,6 +306,17 @@ class LocalParticipant extends Participant<LocalTrackPublication> {
     await room.engine.sendDataPacket(packet);
   }
 
+  @override
+  EncryptionType get encryptionType {
```
davidzhao, line 310: a bit of comments for this method would be helpful. or maybe rename to `firstTrackEncryptionType` to be clear?

lib/src/participant/participant.dart
```diff
@@ -1,20 +1,11 @@
 import 'package:collection/collection.dart';
+import 'package:livekit_client/livekit_client.dart';
```
davidzhao, line 2: TIL that you could do this without worrying about circular dependencies.

lib/src/participant/remote.dart
```diff
@@ -42,6 +43,17 @@ class RemoteParticipant extends Participant<RemoteTrackPublication> {
     updateFromInfo(info);
   }
 
+  @override
+  EncryptionType get encryptionType {
```
davidzhao, line 47: the implementation appears to be the same between local and remote. Could this be defined in participant.dart?

davidzhao (approved):
awesome work! just a couple of nits here and there.

question: how did e2ee.worker.dart.js get generated? namely, for client applications integrating this library, what do they have to do?

cloudwebrtc:
okay, I added the build steps and source code for e2ee.worker.js.  https://github.com/livekit/client-sdk-flutter/pull/226/commits/c3ab41a7aec14b587fecb1844b8b47765b123ea5

lib/src/e2ee/key_provider.dart
```diff
@@ -0,0 +1,84 @@
...
+  final KeyManager _keyManager;
+  @override
+  KeyManager get keyManager => _keyManager;
+
+  Uint8List? get sharedKey => _sharedKey;
+
+  BaseKeyProvider(this._keyManager, this.options);
+
+  static Future<BaseKeyProvider> create({
+    bool sharedKey = false,
```
lukasIO, line 38: shouldn't the sharedKey value be `true` for the default key provider?

lib/src/e2ee/key_provider.dart
```diff
@@ -0,0 +1,86 @@
+import 'dart:typed_data';
+
+import 'package:flutter_webrtc/flutter_webrtc.dart' as rtc;
+
+const defaultRatchetSalt = 'LKFrameEncryptionKey';
+const defualtMagicBytes = 'LK-ROCKS';
```
lukasIO, line 6: nit: typo

example/lib/pages/connect.dart
```diff
@@ -93,6 +103,13 @@ class _ConnectPageState extends State<ConnectPage> {
 
       // Create a Listener before connecting
       final listener = room.createListener();
+      E2EEOptions? e2eeOptions;
+      if (_e2ee) {
+        final keyProvider = await BaseKeyProvider.create(sharedKey: true);
```
lukasIO, line 108: maybe our default implementation doesn't even need `sharedKey` to be configurable, but rather have it set in the constructor.  

that would allow us to _only_ have `keyProvider.setKey(myKey)` as an API for the default implementation and users could create their own more complex ones by inheriting from the base. 
What I'm suggesting is having `ExternalKeyprovider` additionally that inherits from `BaseKeyProvider` but only exposes a single `setKey` method.

lib/src/e2ee/key_provider.dart
lukasIO: also: maybe it makes sense for the lk-server to tell us what magic bytes it's using?
1. we wouldn't have to hard-code it across clients
2. server implementations could change them over time

cc @boks1971

lukasIO (approved):
🔥

lib/src/e2ee/key_provider.dart
boks1971: Yeah, that would be good. It can be random per session/track too. Where should be notify it? In `TrackInfo`?

lukasIO: `TrackPublishedResponse` maybe, not sure. I guess making it configurable per track would mean some additional wrangling on the client side. Any particular scenarios you have in mind for setting it per track rather than session wide?

boks1971: No reason for track wide. Session wide should be fine. Just going with the principles of as granular/as flexible as possible.

Not sure about `TrackPublishedResponse`. The injection happens in down track path. So, subscribe only clients also will need it.

Also, my `TrackInfo` suggestion is bad because it is a publisher side stuff.

Maybe, a separate message for this?

PR #231, opened 2023-01-02
Fix Audio output list is empty in android

moham96:
This used to throw a state error because the audio output list is empty under android ( or at least my android device)


cloudwebrtc (approved):
lgtm

PR #238, opened 2023-02-10
Fix: re-publish tracks after re connect

cloudwebrtc:
bug fixes:
* Fix RtpSender.getParameters error reporting in Firefox.
* Fix mediaStreamTrackSetEnable exception
* Correctly wait for `publisher.pc.state.isConnected()` for `restartConnection();`
* Don't handle fastConnectOptions after reConnected.
* Fix sendSyncState timing when reconnecting event is fired.

### DIFF
--- a/lib/src/core/engine.dart
+++ b/lib/src/core/engine.dart
@@ -550,7 +550,8 @@ class Engine extends Disposable with EventsEmittable<EngineEvent> {
           ].contains(reason);
     }
 
-    if (_connectionState == ConnectionState.reconnecting && !fullReconnect) {
+    if (_restarting ||
+        (_connectionState == ConnectionState.reconnecting && !fullReconnect)) {
       logger.fine('[$objectId] Already reconnecting...');
       return;
     }
@@ -645,7 +646,6 @@ class Engine extends Disposable with EventsEmittable<EngineEvent> {
     _restarting = true;
     await publisher?.dispose();
     publisher = null;
-    _hasPublished = false;
 
     await subscriber?.dispose();
     subscriber = null;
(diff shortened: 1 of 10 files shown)


lib/src/core/engine.dart
```diff
@@ -657,7 +666,23 @@ class Engine extends Disposable with EventsEmittable<EngineEvent> {
       fastConnectOptions: fastConnectOptions,
     );
 
+    bool publisherConnected =
+        publisher?.pc.connectionState?.isConnected() ?? false;
+
+    if (publisher != null && !publisherConnected) {
+      logger.warning('restartConnection: Waiting for publisher to connect...');
+      await events.waitFor<EnginePublisherPeerStateUpdatedEvent>(
```
davidzhao, line 674: I think there's a bit of a race condition here. if peer connection is connected between the previous check and here. we would missed it and have thrown a timeout?

very small window I know. I think similar issues may exist elsewhere too.

davidzhao (approved):

lib/src/core/engine.dart
cloudwebrtc: It should not be here, the pc has been disposed of before the re-connection.
``` dart
await publisher?.dispose();
publisher = null;
......
if (publisher != null && !publisherConnected) {
```

davidzhao: yes.. but I think `connect` will recreate it and establish the connection?

anyways, here it seems like the connection would most definitely take longer than it takes this function to execute the next line.. but more wondering the general issue of using `events.waitFor` while the conditions have become true before that call is made.

PR #241, opened 2023-02-22
fix: Fix renderer issue for Safari/Firefox.

cloudwebrtc:
Bug description (only reproduces in Safari):
After muting the local video track, the remote video track will no longer be displayed.

Not displaying fixed, but causing UI flickering. Still under investigation.


lukasIO (approved):
lgtm!

PR #250, opened 2023-03-07
feat: Forward leave reason of disconnected events.

cloudwebrtc:
implementation  for https://github.com/livekit/protocol/pull/139

### DIFF
--- a/lib/src/types/internal.dart
+++ b/lib/src/types/internal.dart
@@ -1,0 +1,11 @@
+import 'package:meta/meta.dart';
+
+@internal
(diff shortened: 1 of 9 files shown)


lib/src/types/internal.dart
```diff
@@ -0,0 +1,11 @@
+import 'package:meta/meta.dart';
+
+@internal
+enum InternalDisconnectReason {
```
davidzhao, line 4: nit: maybe rename to `ClientDisconnectReason`?

davidzhao (approved):

PR #258, opened 2023-03-15
feat: Support setVideoFPS for subscribe.

cloudwebrtc:
https://github.com/livekit/client-sdk-js/pull/484

### DIFF
--- a/lib/src/publication/remote.dart
+++ b/lib/src/publication/remote.dart
@@ -31,6 +31,10 @@ class RemoteTrackPublication<T extends RemoteTrack>
   bool get enabled => _enabled;
   bool _enabled = true;
 
+  /// The current FPS of the track. This is only available for video tracks.
+  int? _fps;
+  int get fps => _fps ?? 0;
+
   lk_models.VideoQuality _videoQuality = lk_models.VideoQuality.HIGH;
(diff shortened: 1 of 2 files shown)


davidzhao (approved):

lib/src/publication/remote.dart
```diff
@@ -31,6 +31,10 @@ class RemoteTrackPublication<T extends RemoteTrack>
   bool get enabled => _enabled;
   bool _enabled = true;
 
+  /// The current FPS of the track. This is only available for video tracks.
```
davidzhao, line 34:
```suggestion
  /// The current desired FPS of the track. This is only available for video tracks that support SVC.
```

lib/src/publication/remote.dart
```diff
@@ -227,6 +231,12 @@ class RemoteTrackPublication<T extends RemoteTrack>
     sendUpdateTrackSettings();
   }
 
+  Future<void> setVideoFPS(int newValue) async {
```
davidzhao, line 234: a bit of comment here would be helpful. Stating that this is setting the desired FPS, server will do its best to return FPS close to this. It's only supported for video codecs that support SVC currently.

Issue #263, opened 2023-04-03
Android screen sharing lags and delays are high


1056824847:
> When sharing the screen, the frame keeps dropping, especially when the static page is converted to a dynamic interface. It is very smooth when testing the camera here. What is the reason?

**This is the live push code** 

```dart
import 'package:bot_toast/bot_toast.dart';
import 'package:dart_jsonwebtoken/dart_jsonwebtoken.dart';
import 'package:flutter_background/flutter_background.dart';
import 'package:flutter_webrtc/flutter_webrtc.dart';
import 'package:get/get.dart';
import 'package:livekit_client/livekit_client.dart';
import 'package:paperless_cross_platform/app/common/utils/log.dart';
import 'package:paperless_cross_platform/app/controller/meeting/logic.dart';
import 'package:paperless_cross_platform/app/controller/meeting/state.dart';
import 'package:paperless_cross_platform/app/controller/rtc_publish/state.dart';
import 'package:paperless_cross_platform/app/controller/system/logic.dart';
import 'package:paperless_cross_platform/app/controller/system/state.dart';
import 'package:paperless_cross_platform/app/websocket/socket_on_send.dart';

/*
 * @Author ZISE
 * @Description //屏幕共享推送
 **/
class PublishLogic extends GetxService {
  final state = PublishState();
  final MeetingLogic meetingLogic = Get.find<MeetingLogic>();
  final MeetingState meetingState = Get.find<MeetingLogic>().state;
  final SystemLogic systemLogic = Get.find<SystemLogic>();
  final SystemState systemState = Get.find<SystemLogic>().state;

  /*
   * @Author ZISE
   * @Description 开启屏幕共享
   * @return
   **/
  void startShareScreen() async {
    String ip = systemState.ip.value.split(r':')[0];
    const roomOptions = RoomOptions(
        adaptiveStream: false,
        dynacast: false);
    state.room = Room(
      roomOptions: roomOptions,
    );
    await state.room?.connect(
      'ws://$ip:7880',
      createToken(meetingState.nickName.value, meetingState.meetingName.value,
          meetingState.userId),
      // roomOptions: roomOptions,
    );
    DesktopCapturerSource? source;
    try {
      if (GetPlatform.isDesktop) {
        List<DesktopCapturerSource> sourceList = await desktopCapturer
            .getSources(types: [SourceType.Screen, SourceType.Window]);
        //找出主屏幕（屏幕一）
        source = sourceList
[...]

PR #270, opened 2023-04-13
Fix resolution/framerate/bitrate issue for publishVideoTrack

cloudwebrtc:
* add maintainResolution setting for publish screen sharing track.
* add CaptureOptions for setCameraEnabled/setMicrophoneEnabled/setScreenShareEnabled
* Fix the low bit rate caused by the inability to use the preset encoding parameters when the simulcast is turned off.
* Upgrade flutter-webrtc to 0.9.26, fix the default h264 profile-level-id in iOS/macOS cannot encode 1080p or above resolution.

P.S.
Not sure if it has some issue with simulcast encoding. but after turning off the simulcast, you can get smooth video streaming. will debug again for Android.

https://github.com/livekit/client-sdk-flutter/issues/263

### DIFF
--- a/lib/src/utils.dart
+++ b/lib/src/utils.dart
@@ -275,14 +275,12 @@ class Utils {
     required List<VideoParameters> presets,
   }) {
     List<rtc.RTCRtpEncoding> result = [];
-    final rids =
-        videoRids.sublist(videoRids.length - presets.length, videoRids.length);
     presets.forEachIndexed((i, e) {
       if (i >= videoRids.length) {
         return;
       }
       final size = dimensions.min();
-      final rid = rids[i];
+      final rid = videoRids[i];
 
       result.add(e.encoding.toRTCRtpEncoding(
         rid: rid,
--- a/lib/src/utils.dart
+++ b/lib/src/utils.dart
@@ -275,12 +275,14 @@ class Utils {
     required List<VideoParameters> presets,
   }) {
     List<rtc.RTCRtpEncoding> result = [];
+    final rids =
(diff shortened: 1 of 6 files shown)


lib/src/utils.dart
```diff
@@ -275,12 +275,14 @@ class Utils {
     required List<VideoParameters> presets,
   }) {
     List<rtc.RTCRtpEncoding> result = [];
+    final rids =
+        videoRids.sublist(videoRids.length - presets.length, videoRids.length);
```
davidzhao, line 279: sorry I wasn't following this algo? In JS I believe we are just looking for the first N presets that match the rids. so if I gave two presets [low, med], it should map to `low: q, med: h`

with this algo, it seems that it would map to `low: h, med: f`. also if I give it 4 presets, this would crash? (i.e. `sublist(-1, 3)`

cloudwebrtc: My understanding is that if there are only two layers then it is full and half, f,h
but if the rid just marks the track layer and is not used to distinguish layer between full, half, and quarter, then I can revert the changes here

davidzhao: it should be q, h if there are only two layers. I think the server side expects it to go from q to f

also the crash seems like it'd happen regardless?

cloudwebrtc: set to h, f seems to get warnings, https://github.com/livekit/livekit/blob/master/pkg/sfu/buffer/videolayerutils.go#L75-L93 but it should not crash. I tested before reverting, and the livekit server did not crash. I have reverted and followed the server settings and no exceptions will be thrown.

Issue #282, opened 2023-05-11
[bug] no sound on ios browser


develocode777:
**Describe the bug**
There is no sound if using a browser on ios. This doesn't happen with an ios build.
**To Reproduce**
This is reproducible with the example app.
**Expected behavior**
Sound should work in browser too.

davidzhao:
This is due to Safari requirement of requiring user interaction before audio could be played back. A workaround is to require user to click on some UI element first (like a button), before attempting to connect to the Room.

develocode777:
This seems to be the case, but interactions in flutter don't count, how can i get a valid user interaction without dart:html package?

develocode777:
Thanks for fixing this! Will this also work in Safari desktop? Sometimes there is the same problem on this platform too.

cloudwebrtc:
I think the same applies to desktop safari

PR #285, opened 2023-05-15
fix: audio play bug for ios safari. 

cloudwebrtc:
close https://github.com/livekit/client-sdk-flutter/issues/282

when detecting that iOS Safari cannot playback audio, will pop up a dialog box to manually activate the playback.

image

### DIFF
--- a/lib/src/track/web/_audio_html.dart
+++ b/lib/src/track/web/_audio_html.dart
@@ -10,7 +10,7 @@ const audioPrefix = 'livekit_audio_';
 
 Map<String, html.Element> _audioElements = {};
 
-void startAudio(String id, rtc.MediaStreamTrack track) {
+Future<dynamic> startAudio(String id, rtc.MediaStreamTrack track) async {
   if (track is! MediaStreamTrackWeb) {
     return;
   }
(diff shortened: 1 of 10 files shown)


lib/src/track/web/_audio_html.dart
```diff
@@ -27,6 +30,15 @@ void startAudio(String id, rtc.MediaStreamTrack track) {
   final audioStream = html.MediaStream();
   audioStream.addTrack(track.jsTrack);
   audioElement.srcObject = audioStream;
+  return audioElement.play();
+}
+
+Future<void> startAllAudioElement() async {
+  for (final element in _audioElements.values) {
+    if (element is html.AudioElement) {
+      await element.play();
```
lukasIO, line 39: in order to work around [this issue](https://stackoverflow.com/questions/9811429/html5-audio-tag-on-safari-has-a-delay/54119854#54119854)
the js client looks at the `state` of a dummy `AudioContext`. 
I guess you could leave out all the `expWebAudioMix` handling for flutter and just use `getNewAudioContext().state === 'running';`

lukasIO (approved):
lgtm apart from the `AudioContext` workaround that might be needed.

lib/src/track/web/_audio_html.dart
cloudwebrtc: yeah, playback delay can be removed after adding AudioContext
https://github.com/livekit/client-sdk-flutter/pull/285/commits/97ce3114fbb1f9bd3058a819c66f587e38f2dfa3

PR #290, opened 2023-05-22
feat: Stats monitor for Track.

cloudwebrtc:
tx stats:

image

rx stats:

image

### DIFF
--- a/example/lib/widgets/participatn_stats.dart
+++ b/example/lib/widgets/participatn_stats.dart
@@ -31,6 +31,7 @@ class _ParticipantStatsWidgetState extends State<ParticipantStatsWidget> {
         setState(() {
           stats['video tx'] = '${event.currentBitrate.toInt()} kpbs';
           event.stats.forEach((key, value) {
+            stats['$key-encoder'] = '${value.encoderImplementation}';
             stats[key] =
                 '${value.frameWidth}x${value.frameHeight} ${value.framesPerSecond?.toDouble()}fps \n qualityLimitationReason: ${value.qualityLimitationReason}';
           });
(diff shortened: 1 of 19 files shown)


example/lib/widgets/participatn_stats.dart
lukasIO, line 1: nit: typo in the file name

lib/src/options.dart
```diff
@@ -157,7 +157,7 @@ class VideoPublishOptions {
   final List<VideoParameters> screenShareSimulcastLayers;
 
   const VideoPublishOptions({
-    this.videoCodec = 'H264',
+    this.videoCodec = 'vp8',
```
lukasIO, line 160: does this change need to be communicated to users in a new release?

lib/src/track/remote/video.dart
```diff
@@ -18,4 +20,82 @@ class RemoteVideoTrack extends RemoteTrack with VideoTrack {
...
+    }
+
+    final stats = await receiver!.getStats();
+    VideoReceiverStats? receiverStats;
+    for (var v in stats) {
+      if (v.type == 'inbound-rtp') {
+        receiverStats ??= VideoReceiverStats();
+        receiverStats.timestamp = v.timestamp;
+        receiverStats.streamId = v.id;
+        receiverStats.jitter ??= getNumValFromReport(v.values, 'jitter');
```
lukasIO, line 62: just out of curiosity: what's the reason why some of the properties are getting assigned directly and some with a null aware assignment?

lib/src/options.dart
davidzhao: good call, we should include this in the release notes. My guess is Duan is changing this for compatibility reasons?

lib/src/track/remote/video.dart
cloudwebrtc: This is because `id` and `timestamp` are in the fixed parameters of StatsReport, and `jitter/packetReceived` are in the kv map, so the parameters are always uncertain.

lib/src/options.dart
cloudwebrtc: Yes, this modification should have nothing to do with this PR, I forgot to revert it after testing.   :)

lukasIO (approved):
lgtm!

PR #307, opened 2023-06-17
Fix incorrect 4:3 preset bitrates

davidzhao:


furkanKotic:
```
static const h480_43 = VideoParameters(
    dimensions: VideoDimensionsPresets.h480_43,
    encoding: VideoEncoding(
      maxBitrate: 300 * 1000,
      maxFramerate: 20,
    ),
  );

  static const h540_43 = VideoParameters(
    dimensions: VideoDimensionsPresets.h540_43,
    encoding: VideoEncoding(
      maxBitrate: 450 * 1000,
      maxFramerate: 25,
    ),
  );
```
Are the bitrate values ​​of the above resolutions correct? According to the table in IBM, 1200 - 1500 kbps is recommended.
https://support.video.ibm.com/hc/en-us/articles/207852117-Internet-connection-and-recommended-encoding-settings 
image

cloudwebrtc (approved):

davidzhao:
> Are the bitrate values ​​of the above resolutions correct? According to the table in IBM, 1200 - 1500 kbps is recommended. https://support.video.ibm.com/hc/en-us/articles/207852117-Internet-connection-and-recommended-encoding-settings image

Yes, these are sane defaults. Note the keyframe interval in the docs that you've linked. Generating a keyframe each second is very inefficient.

Also, please feel free to customize these bitrates as you see fit! These are just suggested values and it's very possible they are not a good fit for your use case.

PR #320, opened 2023-07-13
Clarifying events in flutter example: reacting to participant events for updating UI

talksik:
I ran into issues with implementation and found that some events were not triggered under RoomEvent. Then I found that I needed to listen to the high level `ParticipantEvent` to ensure my UI was accurate when other peers were unpublishing, muting, etc.

This PR hopefully helps with beginners modeling their own code with this example implementation.

- explaining the different listeners and callbacks
- explaining `FastConnectOptions`

Hoping to help more with this process so that the SDK is not a _leaky abstraction._

### DIFF
--- a/lib/src/options.dart
+++ b/lib/src/options.dart
@@ -16,6 +16,9 @@ class TrackOption<E extends Object, T extends Object> {
   const TrackOption({this.enabled, this.track});
 }
 
+/// This will enable the local participant to publish tracks on connect,
+/// instead of having to explicitly publish them.
+/// Defaults to false for all three tracks: microphone, camera, and screen.
 class FastConnectOptions {
   FastConnectOptions({
     this.microphone = const TrackOption(enabled: false),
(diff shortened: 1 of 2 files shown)


lib/src/options.dart
```diff
@@ -16,6 +16,9 @@ class TrackOption<E extends Object, T extends Object> {
   const TrackOption({this.enabled, this.track});
 }
 
+/// This will enable the local participant to publish tracks on connect,
+/// instead of having to explicitly publish them.
+/// Defaults to false for all three tracks: microphone, camera, and screen.
```
cloudwebrtc, line 21:
```suggestion
/// Defaults to false for all three tracks: microphone, camera, and screen.
/// You can also create LocalAudio/VideoTrack on your `PreJoin` page 
/// (preview camera or select audio device), Automatically publish these
/// tracks after the room is connected.
```

cloudwebrtc (approved):
lgtm

Issue #327, opened 2023-08-02
[bug] setSpeakerphoneOn not working


furkanKotic:
**Describe the bug**
Sound output works fine in version 1.3.5, Sound comes from Speaker, it seems to be earpiece by default in version 1.4.0. Sound comes from earpiece.
`Hardware.instance.setSpeakerphoneOn(true);` not working. 
`Hardware.instance.canSwitchSpeakerphone` always returns false.

The example project has the same problem. I think it's a problem with https://github.com/livekit/client-sdk-flutter/pull/312.

**Expected behavior**

The way I want it to work is as follows: If there is a bluetooth headset, the sound should come from here, if not, the should come from the speaker.

**Platform information**

* **Flutter version**: Flutter 3.10.6
* **Plugin version**: livekit_client: ^1.4.0
* **Flutter target OS**: Android

furkanKotic:
@cloudwebrtc 
The `setSpeakerphoneOn `function works now, this problem is solved, but the audio output order is completely messed up.

Previously, when there were wireless headphones, the sound was coming from there, otherwise it was coming from the speaker, if it was not there, the sound was coming from the earpiecer.

Now, if I don't call `setSpeakerphoneOn(true)`, there is never any sound from the speaker, if I set it to true, the bluetooth headset is disabled.

cloudwebrtc:
okay, When you call `setSpeakerphoneOn(false)`, does the sound switch between the earpiece or bluetooth or wired headset correctly? Headphones usually take precedence over earpieces.

furkanKotic:
When I call `setSpeakerphoneOn(false)` it switches to wireless headset.
But if the user doesn't have a headset, it must come from the speaker, so it looks like you need to call` setSpeakerphoneOn(true)` .

Do we call `setSpeakerphoneOn(true)` only when he has no headset? If so, how do we detect that it doesn't have a headset?

davidzhao:
@furkanKotic what if you don't call `setSpeakerphoneOn` ? what is the default behavior in the following cases:

* With headset
* Without headset

furkanKotic:
@davidzhao 

- With headset
The sound is coming from the headset.

- Without headset
The sound is coming from the earphone. not speaker.


1.3.5 version does not have this problem.

cloudwebrtc:
@furkanKotic Can you try this branch?

https://github.com/livekit/client-sdk-flutter/pull/333

furkanKotic:
[...]

PR #333, opened 2023-08-09
fix: set SpeakerPhone on but prefer Bluetooth.

cloudwebrtc:
Revert the changes to around 1.3.5


davidzhao:
Adding DL to review since he's more familiar with audio handling on Android

cloudwebrtc:
Since only reverting changes from 1.3.5 and user testing feedback have been fixed,  so I'm merging it now for the new version release.
https://github.com/livekit/client-sdk-flutter/issues/327#issuecomment-1676872254

Issue #337, opened 2023-08-14
preferCurrentTab should not be defaulted to true


jezell:
As of Chrome 115, this results in you not being able to share your screen or another tab.

jezell:
https://caniuse.com/?search=getDisplayMedia

Pay attention to this note here for preferCurrentTab:

"This feature is non-standard and should not be used without careful consideration."

PR #338, opened 2023-08-15
fix: set preferCurrentTab to false bye default

cloudwebrtc:
close https://github.com/livekit/client-sdk-flutter/issues/337

### DIFF
--- a/lib/src/track/options.dart
+++ b/lib/src/track/options.dart
@@ -102,7 +102,8 @@ class ScreenShareCaptureOptions extends VideoCaptureOptions {
   final bool captureScreenAudio;
 
   /// for browser only, if true, will capture current tab.
-  final bool preferCurrentTab;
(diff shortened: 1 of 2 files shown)


lib/src/track/options.dart
```diff
@@ -102,15 +102,16 @@ class ScreenShareCaptureOptions extends VideoCaptureOptions {
   final bool captureScreenAudio;
 
   /// for browser only, if true, will capture current tab.
-  final bool preferCurrentTab;
+  @Deprecated('Deprecated parameter no longer works in chrome m115+')
```
davidzhao, line 105: I don't think this is deprecated.. what do you mean it no longer works?

davidzhao: I think the problem is that we've defaulted it to true.. the default should be false.

cloudwebrtc: ah, okay.

davidzhao (approved):
lg

PR #349, opened 2023-09-01
Bugfixes with multicodec simulcast

davidzhao:
* No longer require explicit settings in BackupVideoCodec
* Fix crashes when backupCodec isn't set
* Avoid using default settings, instead use the track's publication options
* Correctly set SVC codec layers

### DIFF
--- a/example/android/build.gradle
+++ b/example/android/build.gradle
@@ -24,6 +24,6 @@ subprojects {
     project.evaluationDependsOn(':app')
 }
 
-task clean(type: Delete) {
+tasks.register("clean", Delete) {
     delete rootProject.buildDir
 }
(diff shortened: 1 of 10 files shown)


davidzhao:
There's one remaining issue, which is when AV1 is used, the publisher's connection quality is marked POOR.

cloudwebrtc (approved):
lgtm

cloudwebrtc:
> There's one remaining issue, which is when AV1 is used, the publisher's connection quality is marked POOR. This doesn't happen when no subscriber is in the room.

Yes, I have observed this, but the ConnectionQualityInfo seems to be sent by SFU, is it possible that the server is not calculating the AV1 quality accurately? When I received the POOR, the publisher reported from rtcStats that the total sending bit rate was about 2500kbps.

image
image

PR #350, opened 2023-09-02
Fix SVC codec bitrate computation

davidzhao:
Casing double to int would cause a crash. Previously this path wasn't exercised because we've always hardcoded preset bitrate for SVC


cloudwebrtc (approved):
lgtm

PR #353, opened 2023-09-04
fix: Fixed the bug that e2ee cannot be enabled when room.connect uses…

cloudwebrtc:


lib/src/core/room.dart
```diff
@@ -165,11 +165,12 @@ class Room extends DisposableChangeNotifier with EventsEmittable<RoomEvent> {
     RoomOptions? roomOptions,
     FastConnectOptions? fastConnectOptions,
   }) {
-    if (roomOptions?.e2eeOptions != null) {
+    roomOptions ??= this.roomOptions;
```
davidzhao, line 168: good fix.. in the future I think we should deprecate roomOptions from `connect`. this should really be set in Room constructor.

PR #363, opened 2023-09-15
Improvements for framecryptor

cloudwebrtc:


lib/src/participant/local.dart
```diff
@@ -200,34 +200,36 @@ class LocalParticipant extends Participant<LocalTrackPublication> {
...
-            codec: publishOptions.videoCodec,
-            cid: track.getCid(),
-            enableSimulcastLayers: true),
+          codec: publishOptions.videoCodec,
+          cid: track.getCid(),
+        ),
         lk_rtc.SimulcastCodec(
-            codec: publishOptions.backupCodec!.codec.toLowerCase(),
-            cid: '',
-            enableSimulcastLayers: publishOptions.backupCodec!.simulcast),
```
davidzhao, line 209: these aren't required anymore?

davidzhao (approved):

lib/src/participant/local.dart
cloudwebrtc: When I upgraded the protocol, enableSimulcastLayers seemed to be removed

cloudwebrtc: https://github.com/livekit/protocol/commit/13abb13b71b7461d04be5b651a9938454655bf2c

davidzhao: got it. I guess we don't use it on the server side.

PR #378, opened 2023-10-08
Fix captureScreenAudio conditional

kNoAPP:
**Expected**
When I pass `false` to `captureScreenAudio`, an audio track should not be created/published to the LiveKit room.

**Actual**
Whenever `captureScreenAudio` is not null, an audio track is always created/published to the LiveKit room regardless of if the variable is `false` or not.

**Fix**
Updated the `captureScreenAudio` conditional to default to `false` if the variable is `null`. Otherwise the actual `true`/`false` value is used for the conditional.


cloudwebrtc (approved):
lgtm

PR #382, opened 2023-10-09
Update build.gradle for gradle 8.0.0 namespace

mostcute:
after agp-8.0.0 ,module need to add namespace in build.gradle

https://developer.android.com/build/releases/past-releases/agp-8-0-0-release-notes


cloudwebrtc (approved):
lgtm

PR #387, opened 2023-10-17
fix: low-resolution screen sharing for safari 17.

cloudwebrtc:
related fix on js  https://github.com/livekit/client-sdk-js/pull/889

### DIFF
--- a/lib/src/track/local/local.dart
+++ b/lib/src/track/local/local.dart
@@ -154,6 +154,14 @@ abstract class LocalTrack extends Track {
         if (options.selfBrowserSurface != null) {
           constraints['selfBrowserSurface'] = options.selfBrowserSurface!;
         }
+
+        // Remove resolution settings to avoid capturing low-resolution screen sharing
+        // streams in Safari 17.
+        // related bug for safari 17: https://bugs.webkit.org/show_bug.cgi?id=263015
+        if (lkBrowser() == BrowserType.safari &&
+            lkBrowserVersion().major == 17) {
+          constraints['video'] = true;
+        }
       }
       stream = await rtc.navigator.mediaDevices.getDisplayMedia(constraints);
     } else {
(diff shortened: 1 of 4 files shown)


lib/src/track/local/local.dart
```diff
@@ -154,6 +154,13 @@ abstract class LocalTrack extends Track {
         if (options.selfBrowserSurface != null) {
           constraints['selfBrowserSurface'] = options.selfBrowserSurface!;
         }
+
+        // Remove resolution settings to fix low-resolution screen share on Safari 17.
+        // related bug: https://bugs.webkit.org/show_bug.cgi?id=263015
+        if (lkBrowser() == BrowserType.safari &&
+            lkBrowserVersion().major == 17) {
```
lukasIO, line 161: good thinking to specifically fix it for this Safari version, too. 
I think we should do something similar on the JS side. 
The bugreport in webkit only talks about `width: { max: ` constraints. maybe `width: { ideal: ` would be possible as a workaround for Safari 17. 
I remember it not being a valid workaround in general because `ideal` was causing issues in other Safari versions :/

lukasIO, line 161: good thinking to specifically fix it for this Safari version, too. 
I think we should do something similar on the JS side. 
The bugreport in webkit only talks about `width: { max: ` constraints. maybe `width: { ideal: ` would be possible as a workaround for Safari 17. 
I remember it not being a valid workaround in general because `ideal` was causing issues in other Safari versions :/

cloudwebrtc: just tested using `width: { ideal:`, doesn't seem to fix it, I still get the low resolution screen capture :(
```dart
        if (lkBrowser() == BrowserType.safari &&
            lkBrowserVersion().major == 17) {
          constraints['video'] = {
            'width': {'ideal': options.params.dimensions.width},
            'height': {'ideal': options.params.dimensions.height},
          };
        }
```

lukasIO: thanks for checking. then your fix seems like the best option :)

lukasIO (approved):

davidzhao (approved):
lgtm

PR #391, opened 2023-10-30
handling of incompatible published codecs

cloudwebrtc:
js: https://github.com/livekit/client-sdk-js/pull/911

### DIFF
--- a/lib/src/participant/local.dart
+++ b/lib/src/participant/local.dart
@@ -193,7 +193,7 @@ class LocalParticipant extends Participant<LocalTrackPublication> {
         'Compute encodings with resolution: ${dimensions}, options: ${publishOptions}');
 
     // Video encodings and simulcasts
-    final encodings = Utils.computeVideoEncodings(
+    var encodings = Utils.computeVideoEncodings(
       isScreenShare: track.source == TrackSource.screenShareVideo,
       dimensions: dimensions,
       options: publishOptions,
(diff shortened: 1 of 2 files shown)


lib/src/participant/local.dart
```diff
@@ -250,6 +243,30 @@ class LocalParticipant extends Participant<LocalTrackPublication> {
 
     await track.start();
 
+    String? primaryCodecMime;
+    for (var codec in trackInfo.codecs) {
+      primaryCodecMime ??= codec.mimeType;
+    }
+
+    if (primaryCodecMime != null) {
```
davidzhao, line 251: should only update if track is video

lib/src/utils.dart
```diff
@@ -620,3 +620,11 @@ class ScalabilityMode {
     return 'L${spatial}T${temporal}${suffix ?? ''}';
   }
 }
+
+String mimeTypeToVideoCodecString(String mimeType) {
+  final codec = mimeType.split('/')[1].toLowerCase();
```
davidzhao, line 625: should guard if mimeType isn't valid

davidzhao (commented):
one

lib/src/participant/local.dart
cloudwebrtc: flutter SDK uses a separate function `publishVideoTrack` to publish track, so there should be no need to check `track is video`

davidzhao: ah that makes sense. sorry I missed that.

davidzhao (approved):
awesome!

PR #396, opened 2023-11-07
Simplify backupCodec setting.

cloudwebrtc:


lib/src/options.dart
```diff
@@ -200,16 +201,19 @@ class VideoPublishOptions {
...
   const VideoPublishOptions({
-    this.videoCodec = 'H264',
+    this.videoCodec = defaultVideoCodec,
     this.videoEncoding,
     this.simulcast = true,
     this.videoSimulcastLayers = const [],
     this.screenShareSimulcastLayers = const [],
     this.name,
-    this.backupCodec,
+    this.enableBackupCodec = false,
```
lukasIO, line 215: I think we want to enable it by default

lukasIO (changes requested):

lib/src/options.dart
```diff
@@ -219,7 +223,8 @@ class VideoPublishOptions {
     List<VideoParameters>? videoSimulcastLayers,
     List<VideoParameters>? screenShareSimulcastLayers,
     String? videoCodec,
-    BackupVideoCodec? backupCodec,
+    bool? enableBackupCodec,
```
davidzhao, line 226: I think we ended up settling for having this inside `BackupVideoCodec` type?

cloudwebrtc: Removed `bool? enableBackupCodec`, only use `nullable backupVideoCodec`

davidzhao:
how would someone explicitly disable this?

cloudwebrtc:
okay, updated the BackupVideoCodec structure,   
It will be enabled when backupVideoCodec != null and backupVideoCodec.enabled == true. Do we need to set default for backupCodec? Enabled by default unless the user wants to disable

lukasIO:
yeah, by default it should be enabled unless e2ee is enabled (which automatically needs to disable backup codecs)

davidzhao (approved):
lgtm

Issue #392, opened 2023-10-31
this is an issue with ending screen sharing in iOS.


std-s:
Hi, this is an issue with ending screen sharing in iOS.
When you click the button on the status bar to end screen sharing, how can you monitor the ending action in the app?

cloudwebrtc:
fixed https://github.com/livekit/client-sdk-flutter/pull/398

Issue #242, opened 2023-02-23
how to stop screen share from app 


mohd-afid:
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Platform information**

* **Flutter version**: 
* **Plugin version**:
* **Flutter target OS**: 
* **Flutter target OS version**:
* **Flutter console log**:

PR #398, opened 2023-11-09
fix: set setScreenShareEnabled when detect replaykit state changed

lambiengcode:
I found that the example app is missing the active Broadcast Extension tracking or is disabled. This PR fix the #242 and #392. 

I'll set setScreenShareEnabled is true if detect function `broadcastStarted()` called in `SampleHandler` and `isScreenShareEnabled() == false` or set setScreenShareEnabled is `false` if detect function `broadcastFinished()` called in `SampleHandler` and `isScreenShareEnabled() == false`

```dart
_replayKitChannel.setMethodCallHandler((call) async {
      if (call.method == 'closeReplayKitFromNative') {
        if (!(room.localParticipant?.isScreenShareEnabled() ?? false)) return;

        await room.localParticipant?.setScreenShareEnabled(false);
      } else if (call.method == 'hasSampleBroadcast') {
        if (room.localParticipant?.isScreenShareEnabled() ?? true) return;

        await room.localParticipant?.setScreenShareEnabled(true);
      }
    });
```


cloudwebrtc (approved):
lgtm

cloudwebrtc:
Will create PR for README.md later

Issue #346, opened 2023-08-29
[bug] Error: peerConnection not found


furkanKotic:
**Describe the bug**
The following errors occurs a lot :
```
RemoteAudioTrack.getReceiverStats
FlutterError - Unable to RTCRtpReceiverNative::getStats: Error: peerConnection not found!

Fatal Exception: FlutterError
0  ???                            0x0 RTCRtpReceiverNative.getStats + 45 (rtc_rtp_receiver_impl.dart:45)
1  ???                            0x0 RemoteAudioTrack.getReceiverStats + 105 (audio.dart:105)
2  ???                            0x0 RemoteAudioTrack.monitorStats + 88 (audio.dart:88)
3  ???                            0x0 Track.startMonitor.<fn> + 170 (track.dart:170)
```

Also this error : 
```
Fatal Exception: io.flutter.plugins.firebase.crashlytics.FlutterError: Unable to RTCRtpReceiverNative::getStats: peerConnectionGetStats(): peerConnection is null
       at RTCRtpReceiverNative.getStats(rtc_rtp_receiver_impl.dart:45)
       at RemoteAudioTrack.getReceiverStats(audio.dart:105)
       at RemoteAudioTrack.monitorStats(audio.dart:88)
       at Track.startMonitor.<fn>(track.dart:170)
```

**Platform information**

* **Flutter version**:  Flutter 3.13.1
* **Plugin version**: livekit_client: ^1.4.3
* **Flutter target OS**: Android/iOS

davidzhao:
I think this happens when the user is disconnected, and the stats queue isn't shut down correctly. We'll take a look

afl-dev:
up

PR #399, opened 2023-11-14
fix: fix getStats for remote track.

cloudwebrtc:
close https://github.com/livekit/client-sdk-flutter/issues/346


davidzhao (approved):
this lgtm.

do we want a check to ensure it's all bypassed when peerConnection is null?

Issue #413, opened 2023-11-26
[bug] on Safari the screen sharing does not work. 


patrikheinonen:
**Describe the bug**

Found a bug that on flutter livekit client on web builds on safari the screen sharing doesent work anymore. It prompts you to select a screen to share but nothing gets shared. The os thinks its sharing but livekit doesent. 
 Im not sure if this is related to new safari feeatures

**To Reproduce**

Try screensharing on example app in safari

**Expected behavior**

Screensharing should work.

**Platform information**

Web

* **Flutter version**: 

[✓] Flutter (Channel stable, 3.16.0, on macOS 14.0 23A344 darwin-arm64, locale en-FI)

* **Plugin version**:

livekit_client: ^1.5.3

* **Flutter target OS**: 

Web

* **Flutter target OS version**:

Web

* **Flutter console log**:

PR #415, opened 2023-11-28
Fix safari screen sharing failure.

cloudwebrtc:
Unable to obtain the correct height/width from the safari screen sharing track, which will cause Utils.computeVideoLayers to fail and throw an exception.

close https://github.com/livekit/client-sdk-flutter/issues/413

### DIFF
--- a/lib/src/participant/local.dart
+++ b/lib/src/participant/local.dart
@@ -178,10 +178,9 @@ class LocalParticipant extends Participant<LocalTrackPublication> {
       try {
         // try to use getSettings for more accurate resolution
         final settings = track.mediaStreamTrack.getSettings();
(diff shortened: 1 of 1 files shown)


lib/src/participant/local.dart
```diff
@@ -178,10 +178,10 @@ class LocalParticipant extends Participant<LocalTrackPublication> {
       try {
         // try to use getSettings for more accurate resolution
         final settings = track.mediaStreamTrack.getSettings();
-        if (settings['width'] is int) {
+        if (settings['width'] is int && settings['width'] as int > 0) {
```
davidzhao, line 181: how about combining the two `ifs` into one?

davidzhao (approved):

PR #435, opened 2023-12-22
Rename some parameters.

cloudwebrtc:


lib/src/participant/local.dart
```diff
@@ -457,6 +457,33 @@ class LocalParticipant extends Participant<LocalTrackPublication> {
...
+    final track = super.getTrackPublicationBySid(sid);
+    if (track != null) {
+      return track;
+    }
+    return null;
+  }
+
+  @override
+  LocalTrackPublication? getTrackPublicationBySource(TrackSource source) {
+    final track = super.getTrackPublicationBySid(sid);
```
davidzhao, line 480: hmm how does this work?

lib/src/participant/participant.dart
```diff
@@ -272,6 +267,35 @@ abstract class Participant<T extends TrackPublication>
                 e.kind == lk_models.TrackType.AUDIO));
   }
 
+  // Must be implemented by subclasses.
+  Future<void> unpublishTrack(String trackSid, {bool notify = true});
```
davidzhao, line 271: should this be moved to local participant? what are the benefits of keeping this in the super class?

lib/src/participant/remote.dart
```diff
@@ -78,9 +78,30 @@ class RemoteParticipant extends Participant<RemoteTrackPublication> {
...
+    final track = super.getTrackPublicationBySid(sid);
+    if (track != null) {
+      return track;
+    }
+    return null;
+  }
+
+  @override
+  RemoteTrackPublication? getTrackPublicationBySource(TrackSource source) {
+    final track = super.getTrackPublicationBySid(sid);
```
davidzhao, line 101: also incorrect

lib/src/participant/participant.dart
cloudwebrtc: yeah, moved unpublishTrack, unpublishAllTracks to local participant.

davidzhao (approved):
lg

lib/src/core/room.dart
```diff
@@ -62,9 +62,9 @@ class Room extends DisposableChangeNotifier with EventsEmittable<RoomEvent> {
   RoomOptions get roomOptions => engine.roomOptions;
 
   /// map of SID to RemoteParticipant
-  UnmodifiableMapView<String, RemoteParticipant> get participants =>
-      UnmodifiableMapView(_participants);
-  final _participants = <String, RemoteParticipant>{};
+  UnmodifiableMapView<String, RemoteParticipant> get remoteParticipants =>
```
lukasIO, line 65: As `remoteParticipants` is publicly accessible, I think we'd want to switch to mapping them by `identites` here instead of `sid`. 
cc @davidzhao for confirmation

cloudwebrtc: ah, let me check.

davidzhao: ah yes, great catch!

cloudwebrtc: Resolved https://github.com/livekit/client-sdk-flutter/pull/435/commits/bfe138f0e4d1acb9c91e62a1edb423b6d3653410

PR #436, opened 2023-12-22
Change publishData signature.

cloudwebrtc:


lib/src/participant/local.dart
```diff
@@ -403,16 +403,18 @@ class LocalParticipant extends Participant<LocalTrackPublication> {
...
     final packet = lk_models.DataPacket(
-      kind: reliability.toPBType(),
+      kind: reliable == true
+          ? lk_models.DataPacket_Kind.RELIABLE
+          : lk_models.DataPacket_Kind.LOSSY,
       user: lk_models.UserPacket(
         payload: data,
         participantSid: sid,
-        destinationSids: destinationSids,
+        destinationSids: identities,
```
davidzhao, line 417: this is incorrect. we should not be setting destinationSIds.

davidzhao (changes requested):

davidzhao (approved):

PR #441, opened 2023-12-25
Fix/some bugs reported by users

cloudwebrtc:


lib/src/core/engine.dart
```diff
@@ -823,11 +823,8 @@ class Engine extends Disposable with EventsEmittable<EngineEvent> {
   }) async {
     final previousAnswer =
         (await subscriber?.pc.getLocalDescription())?.toPBType();
-    final previousOffer =
```
davidzhao, line 826: can you clarify why this was removed?

lib/src/widgets/video_track_renderer.dart
```diff
@@ -52,10 +53,18 @@ class VideoTrackRenderer extends StatefulWidget {
 
 class _VideoTrackRendererState extends State<VideoTrackRenderer> {
   rtc.RTCVideoRenderer? _renderer;
+  bool _rendererReady = false;
   EventsListener<TrackEvent>? _listener;
   // Used to compute visibility information
   late GlobalKey _internalKey;
 
+  Future<rtc.RTCVideoRenderer?> _initializeRenderer() async {
```
davidzhao, line 61: seems like it's not possible for _renderer to be null? would be better to reflect that in the method signature.

lib/src/widgets/video_track_renderer.dart
```diff
@@ -52,10 +53,18 @@ class VideoTrackRenderer extends StatefulWidget {
 
 class _VideoTrackRendererState extends State<VideoTrackRenderer> {
   rtc.RTCVideoRenderer? _renderer;
+  bool _rendererReady = false;
```
davidzhao, line 56: if only used for web, could we reflect in variable name? along with a comment it's web-only.

davidzhao (commented):
I only understood parts of it. could you clarify/link to the bugs that this PR is fixing?

lib/src/core/engine.dart
cloudwebrtc: There was an internal exception in flutter-webrtc, so I temporarily removed the code (added from 1.5.5) and added it back after testing
image

davidzhao (approved):

PR #443, opened 2024-01-04
add lost quality (Protocol v11). 

cloudwebrtc:


lib/src/types/other.dart
```diff
@@ -30,6 +30,8 @@ enum ProtocolVersion {
   v7, // Remote unpublish
   v8,
   v9,
+  v10,
```
lukasIO, line 33: just a side note, we should only merge this, when also the v10 feature of sync stream id has been implemented. Otherwise the server will falsely assume that it is supported already.

PR #444, opened 2024-01-04
Use enum VideoQuality/TrackType without exporting PBType.

cloudwebrtc:


lib/src/publication/remote.dart
```diff
@@ -235,11 +235,11 @@ class RemoteTrackPublication<T extends RemoteTrack>
   }
 
   @Deprecated('use setVideoQuality() instead')
```
lukasIO, line 237: if this is deprecated and the PR is part of v2, we can probably remove it here (unless this is removed in the deprecation PR already, then ignore me :) )

lukasIO (approved):
lgtm!

lib/src/publication/remote.dart
cloudwebrtc: yes, he will be deleted in another PR. I changed it here just to not break the dart analyzer.

davidzhao (approved):
nice work!

one question: is it necessary to change it to lower case enums? I think we still use some upper case enums in the code base? Question is if it's worth making users change all of these enums.

cloudwebrtc:
> nice work!
> 
> one question: is it necessary to change it to lower case enums? I think we still use some upper case enums in the code base? Question is if it's worth making users change all of these enums.

Makes sense, but changing to uppercase will cause a dart-style prompt requiring `lowerCamelCase`, unless we change `analysis_options.yaml` to disable lowerCamelCase mandatory requirements.

image

davidzhao:
> Makes sense, but changing to uppercase will cause a dart-style prompt requiring `lowerCamelCase`, unless we change `analysis_options.yaml` to disable lowerCamelCase mandatory requirements.

Got it. I think that would be preferable instead of forcing users to update all the enum references. wdyt?

PR #476, opened 2024-02-29
fix: use latest key index for new rtp nodes

td-famedly:
fixes a bug where if you already have keys before setting up the rtp{sender, receiver} framecryptors, they frames use key index 0 by default


cloudwebrtc (approved):
lg!

PR #481, opened 2024-03-08
Prevent publication type mismatch.

cloudwebrtc:


cloudwebrtc:
Not a valid fix, so closed for now

PR #484, opened 2024-03-12
update to package:web

jezell:
updates to package:web 

see also: 

https://github.com/flutter-webrtc/flutter-webrtc/pull/1546


davidzhao:
Thanks for the PR @jezell. I think @cloudwebrtc is currently investigating this as well

cloudwebrtc:
@jezell I'm migrating dart-webrtc as a whole to web pkg. Only modify the e2ee part, there will be some errors when running.

PR #516, opened 2024-05-19
feat: provide option to skip stop/restartTrack for mute/unmute

td-famedly:
This also fallbacks to the default capture options when null

### DIFF
--- a/lib/src/track/local/local.dart
+++ b/lib/src/track/local/local.dart
@@ -91,11 +91,11 @@ abstract class LocalTrack extends Track {
   /// Mutes this [LocalTrack]. This will stop the sending of track data
   /// and notify the [RemoteParticipant] with [TrackMutedEvent].
   /// Returns true if muted, false if unchanged.
-  Future<bool> mute(bool stopOnMute) async {
+  Future<bool> mute({bool stopOnMute = true}) async {
     logger.fine('LocalTrack.mute() muted: $muted');
     if (muted) return false; // already muted
     await disable();
-    if (!lkPlatformIs(PlatformType.windows) && stopOnMute) {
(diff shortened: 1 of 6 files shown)


lib/src/track/local/local.dart
```diff
@@ -105,10 +105,10 @@ abstract class LocalTrack extends Track {
   /// Un-mutes this [LocalTrack]. This will re-start the sending of track data
   /// and notify the [RemoteParticipant] with [TrackUnmutedEvent].
   /// Returns true if un-muted, false if unchanged.
-  Future<bool> unmute() async {
+  Future<bool> unmute(bool stopOnMute) async {
```
cloudwebrtc, line 108: same as `.mute`
```suggestion
  Future<bool> unmute({bool stopOnMute = true}) async {
```

lib/src/track/local/local.dart
```diff
@@ -91,11 +91,11 @@ abstract class LocalTrack extends Track {
   /// Mutes this [LocalTrack]. This will stop the sending of track data
   /// and notify the [RemoteParticipant] with [TrackMutedEvent].
   /// Returns true if muted, false if unchanged.
-  Future<bool> mute() async {
+  Future<bool> mute(bool stopOnMute) async {
```
cloudwebrtc, line 94: I think optional parameters are needed here to avoid break changes assuming the user is using `.mute` and `.unmute` alone

`{bool stopOnMute = true}`

```suggestion
  Future<bool> mute({bool stopOnMute = true}) async {
```

cloudwebrtc (changes requested):
Everything looks fine. But there are two places to add default parameters to avoid break changes

cloudwebrtc (approved):
lg!

Issue #376, opened 2023-10-05
[bug] iOS AirPods disconnection


furkanKotic:
**Describe the bug**
When a new participant turns their microphone on or off in the room, all participants are disconnected from their AirPods.
fastConnectOptions values when the participant joins the room:

```
fastConnectOptions: FastConnectOptions(
    microphone: const TrackOption(enabled: false),
    camera: const TrackOption(enabled: false),
),
```

Participant microphone on code:
`room!.localParticipant?.setMicrophoneEnabled(true);`

Participant microphone off code:
`room!.localParticipant?.setMicrophoneEnabled(false);`
or
`room!.localParticipant?.unpublishAllTracks();`

When the participant turns the microphone on, off, or stops broadcasting all tracks, everyone's AirPods connection is disconnected.

**Platform information**

* **Flutter version**: Flutter 3.13.6
* **Plugin version**: livekit_client: ^1.5.2
* **Flutter target OS**: iOS real device

furkanKotic:
This problem started to be reported when I upgraded the livekit_client library from v1.3.5 to v1.5.2. So I think it may be caused by the following commits:
https://github.com/flutter-webrtc/flutter-webrtc/pull/1381

davidzhao:
I'm unable to reproduce this. tried with our example app on iOS 17.

damonique-dev:
Our users are able to reproduce this in our TestFlight app in iOS 16.6.1

**Platform information**
  Flutter version: Flutter 3.13.0
  Plugin version: livekit_client: ^1.5.2

afl-dev:
Additionally, when you turn on the microphone (AirPods), the same thing happens and the sound is played from the speaker

furkanKotic:
@davidzhao @cloudwebrtc This problem is also present in your sample project. While there are users in the room, when a new person enters the room, all people already in the room are disconnected from their AirPods.

afl-dev:
up

PR #518, opened 2024-05-24
fix iOS AirPods disconnection.

cloudwebrtc:
close https://github.com/livekit/client-sdk-flutter/issues/376


davidzhao (approved):

furkanKotic:
@cloudwebrtc This problem was solved, but another problem arose. The Airpods problem was solved, but a new problem arose when there were no Airpods.

In my current test, when the Airpods are not connected and the user without any track when entering the room (FastConnectOptions are both false), if he turns on his microphone while the other participants hear it from the Speaker, the sound starts coming from the earpiece.

PR #523, opened 2024-05-31
feat: add voiceIsolation support

td-famedly:


lib/src/track/options.dart
```diff
@@ -269,10 +269,12 @@ class AudioCaptureOptions extends LocalTrackOptions {
     if (!kIsWeb || (kIsWeb && deviceId == null)) {
       constraints['optional'] = <Map<String, dynamic>>[
         <String, dynamic>{'echoCancellation': echoCancellation},
+        <String, dynamic>{'noiseSuppression': noiseSuppression},
+        <String, dynamic>{'autoGainControl': autoGainControl},
+        <String, dynamic>{'voiceIsolation': noiseSuppression},
```
cloudwebrtc, line 274:
```suggestion
        <String, dynamic>{'voiceIsolation': voiceIsolation},
```
It looks like we need to add a parameter here?

cloudwebrtc: okay, `voiceisolation` is actually an enhanced version of `noiseSuppression`. so I think we don't need a new parameter.
https://w3c.github.io/mediacapture-extensions/#voiceisolation-constraint

cloudwebrtc (approved):
lgtm!

Issue #529, opened 2024-06-08
[Question] When can we expect transcription handling in this SDK?


bmmseven:
Hi there,

I noticed the new feature to transcribe audio using Deepgram and other suppliers in conjunction with workers on the server and I would like to use it with Flutter as a client.
Unfortunately, I can't see any implementation of the transcription handling in this SDK.
Are you currently implementing this? When can we expect a release with it?

PS: I could try to implement it myself, but I don't have any experience with protobuf.

davidzhao:
We are working on this. it should be here in a couple of weeks

UserJah:
> We are working on this. it should be here in a couple of weeks

Hello! Do you mean support insertable streams?

PR #531, opened 2024-06-11
feat: add Transcription Event.

cloudwebrtc:
close https://github.com/livekit/client-sdk-flutter/issues/529


cloudwebrtc:
Tested to work with speech-to-text

https://github.com/livekit/agents/tree/main/examples/speech-to-text

PR #554, opened 2024-07-12
Updates to readme

davidzhao:
User was getting confused by the lack of mention of Web in the feature table. Since we've reached feature parity across platform, I've removed the feature table


cloudwebrtc (approved):
lgtm!

Issue #547, opened 2024-07-07
[feature] Add attributes to Participant class


ivalkenburg:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
When issuing a room grant to a participant you can define attributes to the participant (map<string, string>). These attributes are visible on the participant in the web hook participant payload, but do not exist in the flutter participant class, only metadata (string). Is it possible to have these attributes available in participant class as map<string, string>?

**Describe alternatives you've considered**

**Additional context**

davidzhao:
yes, participant attributes is a new feature and it'll be available across all of our SDKs soon

bcherry:
@ivalkenburg this was released in [2.2.2](https://github.com/livekit/client-sdk-flutter/releases/tag/v2.2.2) - let us know if you have any feedback on it

PR #558, opened 2024-07-16
Add support for participant attributes

cloudwebrtc:
close https://github.com/livekit/client-sdk-flutter/issues/547


cloudwebrtc:
Passed in e2e test case
https://github.com/livekit/e2e-flutter/blob/main/lib/src/cases/room/participant_attributes.dart

PR #570, opened 2024-08-08
fix: Fixed confusing room options causing settings to overwrite.

cloudwebrtc:
changes:
Remove `connectOptions` from `Room` constructor,
Remove `roomOptions` from `room.connect`
Copy `encoding` from `defaultCameraCaptureOptions/defaultScreenShareCaptureOptions` when `videoPublishOptions.videoEncoding` is null

### DIFF
--- a/example/lib/pages/prejoin.dart
+++ b/example/lib/pages/prejoin.dart
@@ -190,6 +190,14 @@ class _PreJoinPageState extends State<PreJoinPage> {
                 dimensions: const VideoDimensions(1280, 720),
                 encoding: encoding,
               )),
+          defaultScreenShareCaptureOptions: const ScreenShareCaptureOptions(
+              useiOSBroadcastExtension: true,
+              params: VideoParameters(
+                  dimensions: VideoDimensionsPresets.h1080_169,
+                  encoding: VideoEncoding(
+                    maxBitrate: 3 * 1000 * 1000,
+                    maxFramerate: 15,
+                  ))),
           defaultVideoPublishOptions: VideoPublishOptions(
             simulcast: args.simulcast,
             videoCodec: args.preferredCodec,
--- a/example/lib/pages/prejoin.dart
+++ b/example/lib/pages/prejoin.dart
@@ -165,10 +165,10 @@ class _PreJoinPageState extends State<PreJoinPage> {
 
     try {
       //create new room
-      final room = Room();
-
-      // Create a Listener before connecting
-      final listener = room.createListener();
+      var encoding = VideoEncoding(
+        maxBitrate: 5 * 1000 * 1000,
+        maxFramerate: 30,
+      );
 
       E2EEOptions? e2eeOptions;
       if (args.e2ee && args.e2eeKey != null) {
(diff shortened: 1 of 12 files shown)


example/lib/pages/prejoin.dart
```diff
@@ -177,24 +177,19 @@ class _PreJoinPageState extends State<PreJoinPage> {
...
-            videoCodec: args.preferredCodec,
-            backupVideoCodec: BackupVideoCodec(
-              enabled: args.enableBackupVideoCodec,
-            ),
-          ),
+          defaultCameraCaptureOptions: CameraCaptureOptions(
+              maxFrameRate: 30,
+              params: VideoParameters(
+                dimensions: const VideoDimensions(1280, 720),
+                encoding: encoding,
```
davidzhao, line 191: is there a way to not set this in example? I think it would be confusing to users why they'd have to set this

lib/src/core/room.dart
```diff
@@ -152,10 +150,10 @@ class Room extends DisposableChangeNotifier with EventsEmittable<RoomEvent> {
     String url,
     String token, {
     ConnectOptions? connectOptions,
-    RoomOptions? roomOptions,
```
davidzhao, line 155: let's leave this here.. otherwise it would break folks apps. feel free to mark it as deprecated if that's the intention

lib/src/participant/local.dart
```diff
@@ -148,6 +148,21 @@ class LocalParticipant extends Participant<LocalTrackPublication> {
     publishOptions ??=
         track.lastPublishOptions ?? room.roomOptions.defaultVideoPublishOptions;
 
+    // if no encoding is specified, use the default encoding
+    if (publishOptions.videoEncoding == null) {
```
davidzhao, line 152: is it possible for it to be null here? It seems that room.roomOptions.defaultVideoPublishOptions should always be initialized to something

lib/src/types/video_parameters.dart
```diff
@@ -26,7 +26,7 @@ class VideoParameters implements Comparable<VideoParameters> {
   const VideoParameters({
     this.description,
     required this.dimensions,
-    required this.encoding,
+    this.encoding = VideoEncoding.defaults,
```
davidzhao, line 29: encoding is dimension-specific.. I don't think we should set it to the same across the board. better to make it optional.. and use presets?

lib/src/utils.dart
```diff
@@ -240,9 +240,11 @@ class Utils {
         encoding: VideoEncoding(
           maxBitrate: math.max(
             150 * 1000,
-            (original.encoding.maxBitrate /
-                    (math.pow(scale, 2) *
-                        (original.encoding.maxFramerate / fps)))
+            (original.encoding?.maxBitrate ??
+                    VideoEncoding.defaults.maxBitrate /
```
davidzhao, line 244: I think this is still a problem.. if it's not set.. then users would get stuck with a default encoding?

lib/src/utils.dart
```diff
@@ -274,7 +276,7 @@ class Utils {
     String? codec,
   }) {
     assert(presets.isNotEmpty, 'presets should not be empty');
-    VideoEncoding result = presets.first.encoding;
+    VideoEncoding? result = presets.first.encoding;
```
davidzhao, line 279: preset encoding could never be nil, maybe use `!` here?

lib/src/utils.dart
cloudwebrtc: We should be able to remove `VideoEncoding.defaults`. `encoding` in `original` will not be nil, because `_findAppropriateEncoding` will always assign it an appropriate one
QQ_1723181742704

davidzhao (approved):
lg!

PR #574, opened 2024-08-14
feat: Prepare connection/region pinning.

cloudwebrtc:


lib/src/core/room.dart
```diff
@@ -174,13 +204,65 @@ class Room extends DisposableChangeNotifier with EventsEmittable<RoomEvent> {
...
+      if (_regionUrlProvider == null) {
+        _regionUrlProvider = RegionUrlProvider(url: url, token: token);
+      } else {
+        _regionUrlProvider?.updateToken(token);
+      }
+      // trigger the first fetch without waiting for a response
+      // if initial connection fails, this will speed up picking regional url
+      // on subsequent runs
+      try {
+        await _regionUrlProvider?.fetchRegionSettings();
```
lukasIO, line 221: the comment says `without waiting for a response`, but the code `await`s . 
I think we can remove the `await`

lib/src/support/region_url_provider.dart
```diff
@@ -0,0 +1,139 @@
...
+}
+
+bool isCloudUrl(Uri url) {
+  return url.host.contains('.livekit.cloud') ||
+      url.host.contains('.livekit.run');
+}
+
+String toHttpUrl(String url) {
+  if (url.startsWith('ws')) {
+    return url.replaceAll('ws', 'http');
```
lukasIO, line 129: this would replace `wss://mywsurl.livekit.cloud` with `https://myhttpurl.livekit.cloud`

lib/src/support/region_url_provider.dart
```diff
@@ -0,0 +1,139 @@
...
+  }
+  return url;
+}
+
+String toWebsocketUrl(String url) {
+  if (url.startsWith('http')) {
+    return url.replaceAll('http', 'ws');
```
lukasIO, line 136: same as above

README.md
```diff
@@ -194,6 +194,12 @@ final roomOptions = RoomOptions(
 
 final room = Room();
 
+// if you are using livekit cloud, you can use `prepareConnection`
```
lukasIO, line 197: I think prepareConnection can be used regardless of livekit cloud ? 
Only the region pinning support needs livekit cloud, I think?

lib/src/core/room.dart
```diff
@@ -174,13 +205,65 @@ class Room extends DisposableChangeNotifier with EventsEmittable<RoomEvent> {
...
+      if (_regionUrlProvider == null) {
+        _regionUrlProvider = RegionUrlProvider(url: url, token: token);
+      } else {
+        _regionUrlProvider?.updateToken(token);
+      }
+      // trigger the first fetch without waiting for a response
+      // if initial connection fails, this will speed up picking regional url
+      // on subsequent runs
+      unawaited(_regionUrlProvider?.fetchRegionSettings().then((_) {
+        logger.fine('fetched region settings');
```
lukasIO, line 222: `fetchRegionSettings` doesn't actually update the settings, so we'll need 
```dart
 unawaited(_regionUrlProvider?.fetchRegionSettings().then((settings) {
        _regionUrlProvider.setServerReportedRegions(settings);
```

README.md
cloudwebrtc: Ah, I see.

lukasIO (approved):
lgtm!

PR #594, opened 2024-09-22
fix: Handle disconnect reasons correctly.

cloudwebrtc:
close https://github.com/livekit/livekit/issues/2945

### DIFF
--- a/lib/src/core/engine.dart
+++ b/lib/src/core/engine.dart
@@ -1037,18 +1037,8 @@ class Engine extends Disposable with EventsEmittable<EngineEvent> {
             return;
           }
           await signalClient.cleanUp();
-          if ([
-            DisconnectReason.duplicateIdentity,
-            DisconnectReason.participantRemoved,
-            DisconnectReason.roomDeleted,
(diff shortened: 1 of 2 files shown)


lib/src/core/engine.dart
```diff
@@ -1037,7 +1037,18 @@ class Engine extends Disposable with EventsEmittable<EngineEvent> {
             return;
           }
           await signalClient.cleanUp();
-          await cleanUp();
+          if ([
```
davidzhao, line 1040: disconnect reason actually won't matter. the fields that we should use are:
* [can_reconnect](https://github.com/livekit/protocol/blob/main/protobufs/livekit_rtc.proto#L263)
* [action](https://github.com/livekit/protocol/blob/main/protobufs/livekit_rtc.proto#L265)

reason is just a informational string. what we care about is what the server-indicated action

cloudwebrtc: Solved, tested in e2e-flutter.

davidzhao (approved):
lgtm!

PR #616, opened 2024-10-25
feat: Increase default audio bitrate.

cloudwebrtc:


davidzhao (approved):
thanks!

can we make sure music is the default option selected as well?

PR #624, opened 2024-10-31
fix: more robust mirror rendering detection

holzgeist:
use track settings (see also https://github.com/flutter-webrtc/flutter-webrtc/pull/1695) to detect front/back facing camera


cloudwebrtc (approved):
lg!

PR #638, opened 2024-11-17
feat: audio visualizer.

cloudwebrtc:
image


lib/src/options.dart
```diff
@@ -115,6 +115,9 @@ class RoomOptions {
   /// Options for end-to-end encryption.
   final E2EEOptions? e2eeOptions;
 
+  /// Enable visualizer for audio tracks. Defaults to true.
+  final bool enableVisualizer;
```
bcherry, line 119: can you add more details here to make it clearer what this means? two things come to mind:

1. does it have a negative impact if you turn it on but don't need it? (answer is probably yes)
2. does this actually provide visualization UI automatically? (answer must be no, so point to the other pieces you need to put together to make it work)

bcherry (approved):
this is great - huge effort!

PR #654, opened 2024-12-04
fix: improve reconnect logic.

cloudwebrtc:


bcherry (approved):
Can you include a brief description of the uncovered issue and why this is the right fix?

lib/src/core/signal_client.dart
```diff
@@ -116,8 +118,6 @@ class SignalClient extends Disposable with EventsEmittable<SignalEvent> {
 
       if (_connectivityResult.contains(ConnectivityResult.none)) {
         logger.warning('no internet connection');
-        events.emit(SignalDisconnectedEvent(
```
cloudwebrtc, line 119: There is no need to emit a SignalDisconnectedEvent event here, because connect or resumeConnect/restartConnect will catch the ConnectException and take the correct follow-up action.

PR #675, opened 2024-12-31
Fix side effects for stop remote track.

cloudwebrtc:


std-s:
I tried this version, and when I turned off the local microphone, other people in the room saw that I was disconnected and couldn't hear any sound.

PR #682, opened 2025-01-12
feat: RPC.

cloudwebrtc:
example app https://github.com/livekit-examples/flutter-examples/tree/main/packages

image


bcherry:
@cloudwebrtc see in linear, we also have decided to move the method handlers to room so they can be registered before or after connect. since this hasn't merged yet, no need to maintain deprecated versions on local participant

praveen-nuna:
@bcherry Do you know when RPC be officially supported ?

bcherry:
@praveen-nuna we're actively working on it, should land within a week or two

lib/src/core/room.dart
```diff
@@ -119,6 +121,16 @@ class Room extends DisposableChangeNotifier with EventsEmittable<RoomEvent> {
...
   final Map<String, DateTime> _transcriptionReceivedTimes = {};
 
+  // RPC Handlers
+  final Map<String, RpcRequestHandler> rpcHandlers = {};
+
+  // RPC Pending Acks
+  final Map<String, Function(String participantIdentity)> rpcPendingAcks = {};
+
+  // RPC Pending Responses
+  final Map<String, Function(String? payload, RpcError? error)>
```
lukasIO, line 131: should these all be prefixed with `_` to mark them as private?

lib/src/core/room.dart
```diff
@@ -1129,3 +1143,277 @@ extension RoomHardwareManagementMethods on Room {
...
+  }
+
+  /// Initiate an RPC call to a remote participant.
+  /// @param destinationIdentity - The `identity` of the destination participant
+  /// @param method - The method name to call
+  /// @param payload - The method payload
+  /// @param responseTimeoutMs - Timeout for receiving a response after initial connection
+  /// @returns A promise that resolves with the response payload or rejects with an error.
+  /// @throws Error on failure. Details in `message`.
+  Future<String> performRpc(PerformRpcParams params) async {
```
lukasIO, line 1370: I think the current design has the handler registration on room level, but `performRpc` is still on local_participant, right? 
cc @bcherry

lib/src/core/room.dart
cloudwebrtc: makes sense, yeah, it should be private

lib/src/core/room.dart
bcherry: oh yeah that's correct

bcherry (changes requested):
this is great

bcherry (approved):

PR #695, opened 2025-01-28
fix: tighten VideoTrackRenderer Widget

holzgeist:
This allows external users to e.g. apply rounded edges. Without this patch, VideoTrackRenderer will expand to its available space with no way (known to me..) for parent widgets to know how large the video actually is.

depends on https://github.com/flutter-webrtc/flutter-webrtc/pull/1754 (and https://github.com/flutter-webrtc/webrtc-interface/pull/30)

### DIFF
--- a/lib/src/widgets/video_track_renderer.dart
+++ b/lib/src/widgets/video_track_renderer.dart
@@ -65,6 +65,9 @@ class VideoTrackRenderer extends StatefulWidget {
   final rtc.RTCVideoRenderer? cachedRenderer;
   final bool autoDisposeRenderer;
 
+  /// wrap the video view in a Center widget (if [fit] is [VideoViewFit.contain])
+  final bool autoCenter;
+
   const VideoTrackRenderer(
     this.track, {
     this.fit = VideoViewFit.contain,
@@ -72,6 +75,7 @@ class VideoTrackRenderer extends StatefulWidget {
     this.renderMode = VideoRenderMode.texture,
     this.autoDisposeRenderer = true,
     this.cachedRenderer,
+    this.autoCenter = true,
     Key? key,
   }) : super(key: key);
 
@@ -83,6 +87,7 @@ class _VideoTrackRendererState extends State<VideoTrackRenderer> {
   rtc.VideoRenderer? _renderer;
   // for flutter web only.
   bool _rendererReadyForWeb = false;
+  double? _aspectRatio;
   EventsListener<TrackEvent>? _listener;
   // Used to compute visibility information
   late GlobalKey _internalKey;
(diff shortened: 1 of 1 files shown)


holzgeist:
tests fail due to required changes in flutter-webrtc and webrtc-interface not being merged yet

cloudwebrtc:
This PR will be merged after the new flutter-webrtc is released

cloudwebrtc:
I tested this PR and found that it broke the [fit setting](https://github.com/livekit/client-sdk-flutter/blob/main/lib/src/widgets/video_track_renderer.dart#L54) in VideoTrackRenderer.

when I publish a video track from the portrait viewfinder on a mobile

Original code
RTCVideoViewObjectFitContain

image

RTCVideoViewObjectFitCover
image


Modification from this PR

RTCVideoViewObjectFitContain
image

RTCVideoViewObjectFitCover
image

As shown in the screenshot, PR broke the centering of widgets and the rendering of the screen when covering

holzgeist:
oh right, I didn't test with the cover option. All my sizing should be ignored in cover option as the widget will fill up all available space anyway. I'll push a fix soon

holzgeist:
Hi @cloudwebrtc 

what's the reason for closing this? I added the fix you requested. (Though now I see that I didn't ping you about it..)

cloudwebrtc:
I think this PR breaks the existing layout, this is the main branch layout
image
image

This branch causes the position of the video widgets to shift, the video view is always docked in the upper left corner.
image
image

You can try dragging the window and changing the video size, and the video should always be displayed in the center.

holzgeist:
oh, sorry about that! I did test both options with my app, but not with the example app. I'll fix it and ping you here to reopen the PR (if I get it to work :) )

holzgeist:
It works with the example app now (tested on native linux and web).

The initial issue was that flutter-webrtc wraps the whole view in a center, preventing any "insights" into the actual size. My PR "removes" that Center by wrapping in a LayoutBuilder and SizedBox.
Our app makes use of the original size to add rounded edges and then wraps in a center *again*.
The example app (and very likely other users) don't do this, so I introduced a `autoCenter` parameter to wrap the result in a center. Not for cover fit though, it doesn't matter in that case as all the available space is filled anyway

holzgeist:
@cloudwebrtc should I open another PR or can you reopen this one? It looks like I can't reopen it

holzgeist:
The build failure for android does not seem to be connected to my changes, the PR should be ready for review now

holzgeist:
Hi @cloudwebrtc 
Is there something that's missing in the PR or is it in your queue?
No hurries though if you don't need anything from my side :)

cloudwebrtc (approved):
tested on linux desktop, it seems works fine.

PR #701, opened 2025-02-05
Emit a final empty AudioVisualizerEvent after track stops

bcherry:
This makes it easy for consumers to zero themselves out on mute, etc. An empty array should be safe, consumers will automatically fill in the zero values based on their expected size.


cloudwebrtc (approved):
lg!

PR #718, opened 2025-03-01
feat: Visualizer for web.

cloudwebrtc:
Changes:
- removed `enableVisualizer` from RoomOptions.
- added/exported `AudioVisualizer` and `AudioVisualizerOptions` APIs form SDK.
   Similar to JS, `AudioVisualizer` can be created directly in the component
   https://github.com/livekit/client-sdk-flutter/blob/feat/audio-visualizer-for-web/example/lib/widgets/sound_waveform.dart#L68-L93
-  support `barCount`/`centeredBands` configuration in `AudioVisualizerOptions`
- Audio visualizer for flutter web.
image


lib/src/track/visualizer.dart
```diff
@@ -0,0 +1,25 @@
...
+class VisualizerOptions {
+  final bool centeredBands;
+  final int barCount;
+  const VisualizerOptions({
+    this.centeredBands = true,
+    this.barCount = 7,
+  });
+}
+
+abstract class Visualizer extends DisposableChangeNotifier
```
lukasIO, line 18: the events are called `AudioVisualizerEvents`, maybe it makes sense to call the Visualizer also `AudioVisualizer` for clarity? 

JS's naming is explicit with `BarVisualizer` as we might add more/different visualizers down the road.

lib/src/track/visualizer_web.dart
```diff
@@ -0,0 +1,168 @@
...
+            var element = tmp.toDart[i];
+            frequencies[i] = element;
+          }
+          frequencies = frequencies.sublist(
+              options.loPass!.toInt(), options.hiPass!.toInt());
+
+          final normalizedFrequencies =
+              normalizeFrequencies(frequencies); // is this needed ?
+          final chunkSize = (normalizedFrequencies.length / (bands + 1))
+              .ceil(); // we want logarithmic chunking here
```
lukasIO, line 58: the comment here was intended of more as a `TODO` so that the bins would more accurately reflect how humans hear. I think we can remove the comment for as long as we don't have that.

shared_swift/LiveKitPlugin.swift
```diff
@@ -27,7 +27,8 @@ import Combine
 @available(iOS 13.0, *)
 public class LiveKitPlugin: NSObject, FlutterPlugin {
 
-    var processers: Dictionary<Track, Visualizer> = [:]
+    var processers: Dictionary<String, Visualizer> = [:]
```
lukasIO, line 30:
```suggestion
    var processors: Dictionary<String, Visualizer> = [:]
```

bcherry (approved):

PR #723, opened 2025-03-06
feat: VideoRenderer support cache renderer.

kingiol:
VideoRenderer display in Screen
When UI Layout changed, then video become black, then display new stream.

So i add a cache renderer, keep video stream side the video renderer, avoid VideoRenderer recreate a new renderer.


cloudwebrtc (approved):
lg

PR #728, opened 2025-03-10
Fix: reconnect immediately when disconnect.

kingiol:
When server is disconnected, check if can reconnect.
if so, then connect immediately


cloudwebrtc:
I think we can only `re-connect` when the server explicitly returns `LeaveRequest_Action.RECONNECT`

kingiol:
You are true, but this can connect fast to server as soon as possible

PR #729, opened 2025-03-13
feat: data stream.

cloudwebrtc:


lib/src/participant/local.dart
```diff
@@ -1146,3 +1151,183 @@ extension RPCMethods on LocalParticipant {
...
+      info: info,
+      onClose: () {
+        cancelFun?.call();
+      },
+    );
+
+    return writer;
+  }
+
+  Future<Map<String, String>> sendFile(
```
lukasIO, line 1272: I recently added also the `streamBytes` API to JS. 
This adds consistency across other SDKs and allows for `sendFile` to be a light wrapper around `streamBytes`. see https://github.com/livekit/client-sdk-js/commit/c0c5642ede2d7e202e9d03916d74fb6980d4d48f

lib/src/data_stream/stream_writer.dart
```diff
@@ -0,0 +1,108 @@
...
+}
+
+class TextStreamWriter extends BaseStreamWriter<String, TextStreamInfo> {
+  TextStreamWriter(
+      {required super.writableStream,
+      required super.info,
+      required super.onClose});
+}
+
+class BinaryStreamWriter extends BaseStreamWriter<Uint8List, ByteStreamInfo> {
```
lukasIO, line 38: rename to `ByteStreamWriter` (I had this wrong in the JS implementation, sorry!)

lib/src/data_stream/stream_writer.dart
```diff
@@ -0,0 +1,108 @@
...
+  final StreamWriter<T> writableStream;
+  Function()? onClose;
+
+  final InfoType info;
+
+  BaseStreamWriter(
+      {required this.writableStream, required this.info, this.onClose});
+
+  Future<void> write(T chunk) async {
+    return writableStream.write(chunk);
```
lukasIO, line 22: we also need an implementation for `ByteStreamWriter` where the `write` method automatically chunks oversized byte payloads to a max size kStreamChunkSize

lib/src/data_stream/stream_writer.dart
```diff
@@ -83,26 +80,32 @@ class StreamWriterImpl implements StreamWriter<String> {
...
-  if (text.length <= chunkSize) {
-    chunks.add(Uint8List.fromList(utf8Bytes));
-    return chunks;
-  }
-  for (var i = 0; i < utf8Bytes.length; i += chunkSize) {
-    final end = i + chunkSize;
-    if (end > utf8Bytes.length) {
-      chunks.add(Uint8List.fromList(utf8Bytes.sublist(i)));
-      break;
+  List<Uint8List> splitUTF8List(T chunk, int chunkSize) {
```
lukasIO, line 87: this method doesn't actually split at utf8 characters. 

there's an edge case where a multi byte utf-8 character (e.g. an emoji) is at the exact end of the stream chunk length.

So this multi-byte character would get split in the middle. I think we still need to avoid that. See https://github.com/livekit/client-sdk-js/blob/main/src/room/utils.ts#L629-L652

cloudwebrtc: thanks for the tip and solution. just resolved and passed the unit test for multi-byte string。

lukasIO (approved):
lgtm 🚀

PR #772, opened 2025-04-29
Expose Agent Track Amplitude Levels

CatalinVoss:
This allows users of the flutter SDK to subscribe to the agent amplitude levels.

```
void _handleAudioLevelChange() {

    final audioLevels = room!.remoteTrackAudioLevels;
    final maxLevel = audioLevels.values.isNotEmpty
        ? audioLevels.values.reduce((a, b) => a > b ? a : b)
        : 0.0;
    // maxLevel now gives us the normalized amplitude of the agent track
}

// Note that the Room object itself is a ChangeNotifier. We implemented the logic
// so that the Engine calls a callback (_handleEngineAudioLevelUpdate) on the
// Room instance every 50ms. Inside that handler, the Room updates its
// remoteTrackAudioLevels map and, crucially, calls notifyListeners() on itself.
room.addListener(_handleAudioLevelChange);
```

I really want to get actual PCM data, but that seems more complicated

### DIFF
--- a/lib/src/core/engine.dart
+++ b/lib/src/core/engine.dart
@@ -17,11 +17,11 @@
 import 'dart:async';
 
 import 'package:flutter/foundation.dart';
+import 'package:meta/meta.dart';
 
 import 'package:collection/collection.dart';
 import 'package:connectivity_plus/connectivity_plus.dart';
 import 'package:flutter_webrtc/flutter_webrtc.dart' as rtc;
-import 'package:meta/meta.dart';
 
 import 'package:livekit_client/livekit_client.dart';
 import '../extensions.dart';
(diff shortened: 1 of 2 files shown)


CatalinVoss:
Agh. Looks like this works on web but not device :/

CatalinVoss:
@cloudwebrtc pointed me to instead use AudioVisualizer and set barCount = 1 -- see [`audio_visualizer.dart`](https://github.com/livekit/client-sdk-flutter/blob/main/lib/src/track/audio_visualizer.dart#L14)

I now have a better implementation that works cross-platform following something similar to 

[`sound_waveform.dart`](https://github.com/livekit/client-sdk-flutter/blob/main/example/lib/widgets/sound_waveform.dart#L73-L74)

The only challenge is that the default smoothening is too aggressive for our needs. I will close this PR and open a new one that exposes the smoothening params at least for web.

CatalinVoss:
Superseded by https://github.com/livekit/client-sdk-flutter/pull/774

PR #774, opened 2025-04-30
Fix smoothTransition option issues

CatalinVoss:
- Web: uses `smoothingTimeConstant` within `AudioAnalyserOptions` instead
- Android: fixes missing assignment


cloudwebrtc (approved):
lg

PR #780, opened 2025-05-07
fix: mitigate reconnect loop

holzgeist:
This PR mitigates a reconnect loop that happens via following path:
- a reconnect takes the route through `restartConnection` https://github.com/livekit/client-sdk-flutter/blob/v2.4.6/lib/src/core/engine.dart#L862 (e.g. due to `sendSimulateScenario(fullReconnect: true)`)
- `_setUpSignalListeners` https://github.com/livekit/client-sdk-flutter/blob/v2.4.6/lib/src/core/engine.dart#L888 sets up a callback that will initiate negotiation under certain circumstances (https://github.com/livekit/client-sdk-flutter/blob/v2.4.6/lib/src/core/engine.dart#L980)
- if connection is fast enough (or it is enforced? I'm not 100% sure about the details), the connection of the publisher is setup (https://github.com/livekit/client-sdk-flutter/blob/v2.4.6/lib/src/core/engine.dart#L455) *before* we wait for the event here https://github.com/livekit/client-sdk-flutter/blob/v2.4.6/lib/src/core/engine.dart#L902-L905
- The event won't come again, triggering a timeout, a reconnect and the loop start from the beginning

I extracted the code from `sendDataPacket` into a function that checks if a publisher is connected before attempting to reconnect and reuse it in `restartConnection`

### DIFF
--- a/lib/src/core/engine.dart
+++ b/lib/src/core/engine.dart
@@ -285,22 +285,7 @@ class Engine extends Disposable with EventsEmittable<EngineEvent> {
     if (_subscriberPrimary) {
       // make sure publisher transport is connected
 
-      if ((await publisher?.pc.getConnectionState())?.isConnected() != true) {
-        logger.fine('Publisher is not connected...');
-
-        // start negotiation
-        if (await publisher?.pc.getConnectionState() !=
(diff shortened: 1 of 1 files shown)


cloudwebrtc (approved):
lg!

PR #788, opened 2025-05-22
fix: sif detection for e2ee

td-famedly:


cloudwebrtc (approved):
lg!

td-famedly:
@cloudwebrtc it looks like the ci is being flaky, can you rerun it please?

cloudwebrtc:
I tried to rerun CI several times, but it still failed. I can compile successfully when running locally, but occasionally I get similar errors. This should have nothing to do with the current code. It may be that the flutter sdk has an error in syncing a file.

PR #814, opened 2025-06-26
fix: update camera `deviceId` when track is null

yash-garg:
If there is no `LocalVideoTrack` while we are in the call, `setVideoInputDevice` will skip the device selection as a whole since `track` is `null`.

This will result in a wrong device being used to create a track.

```mermaid
flowchart TD
  A["Join room with video OFF (X)"] --> B[No video track published]
  B --> C["Select new camera device (Y)"]
  C --> D{Is video track available?}
  D -- No --> E[track is null, switchCamera skipped]
  E --> F[roomOptions not updated]
  F --> G[Turn on video]
  G --> H["New track uses old device (X)"]
  H --> I[Camera shows wrong input]
```

### DIFF
--- a/lib/src/core/signal_client.dart
+++ b/lib/src/core/signal_client.dart
@@ -315,6 +315,7 @@ class SignalClient extends Disposable with EventsEmittable<SignalEvent> {
       case lk_rtc.SignalResponse_Message.subscribedQualityUpdate:
         events.emit(SignalSubscribedQualityUpdatedEvent(
           trackSid: msg.subscribedQualityUpdate.trackSid,
+          // ignore: deprecated_member_use_from_same_package
           subscribedQualities: msg.subscribedQualityUpdate.subscribedQualities,
           subscribedCodecs: msg.subscribedQualityUpdate.subscribedCodecs,
         ));


lib/src/core/room.dart
```diff
@@ -1097,17 +1097,17 @@ extension RoomHardwareManagementMethods on Room {
...
-    if (track == null) return;
-    if (selectedVideoInputDeviceId != device.deviceId) {
+
+    // Always update roomOptions so future tracks use the correct device
+    engine.roomOptions = engine.roomOptions.copyWith(
+      defaultCameraCaptureOptions: roomOptions.defaultCameraCaptureOptions
+          .copyWith(deviceId: device.deviceId),
+    );
+
+    if (track != null && selectedVideoInputDeviceId != device.deviceId) {
```
td-famedly, line 1107: if switch camera fails, you probably want to reset the `engine.roomOptions` to the previous state. 

You might also have to do the same for other types of devices

td-famedly (changes requested):

cloudwebrtc (approved):

PR #819, opened 2025-07-08
fix(web, firefox): backup old values before overwriting them

holzgeist:
I'm not sure if the comment about disabling layers in Firefox is still accurate.

In any case, this PR fixes an issue where quality would not get back up anymore once higher quality levels were disabled because the bitrate/framerate/scale values were overwritten in a shallow copy of the encoding

This PR introduces a backup map to properly copy and restore the modified values


td-famedly:
@cloudwebrtc looks like this needs a force merge as well. Thank you!

cloudwebrtc:
hey @td-famedly, This line forces the non-null check to be ignored, which may cause a crash. I think it may need to be modified.

https://github.com/livekit/client-sdk-flutter/pull/819/files#diff-48afd7e75035607dd21e7c35577a8c13949ebe809f6f23529661c6003a1ffb9aR456

holzgeist:
@cloudwebrtc fixed by falling back to the current encoding if there is no backup

holzgeist:
I had to run `dart format` on the file to make CI happy. It did quite some changes unfortunately..
If someone reformats `main`, I'll merge/rebase to make the diff nicer :)

cloudwebrtc:
lg!

PR #842, opened 2025-08-01
Fix linux build error

RBT22:
I ran into a build issue on Linux when using this package. It relies on C++17 features, but without explicitly setting the C++ standard in the package's CMakeLists.txt, the build system was defaulting to an older version.


cloudwebrtc (approved):
lgtm!

PR #867, opened 2025-09-07
Fix publications' track property initialization

hiroshihorie:
The constructors of `LocalTrackPublication` and `RemoteTrackPublication` relied on setting the `track` property during initialization. However, because `updateTrack()` was an un-awaited async method, the `track` property was null immediately after construction, which caused confusion.


pitazzo:
hi @hiroshihorie 
since this PR was merged, we're receiving always null tracks inside the remote's participant video publications. Did this introduce any change on how tracks should be received?

PR #873, opened 2025-09-13
feat: Data Packet Cryptor.

cloudwebrtc:


lib/src/e2ee/e2ee_manager.dart
```diff
@@ -117,6 +121,11 @@ class E2EEManager {
             }
           }
         });
+      if (_encryptionEnabled && room.roomOptions.encryption != null) {
+        _dataPacketCryptor ??=
```
lukasIO, line 125: I think the problem with this is that it only sets up the dataPacketCryptor if `roomOptions.encryption` is set. 

However the behaviour we'd want is to support _de_cryption also when `e2ee` has been set on the room options instead of `encryption`. 
If `encryption` is set, then also _en_cryption for data channels will be performed, but we'll need to make sure that the decryption path works for both cases.

lib/src/options.dart
```diff
@@ -110,8 +110,11 @@ class RoomOptions {
   final bool stopLocalTrackOnUnpublish;
 
   /// Options for end-to-end encryption.
+  @Deprecated('Use encryption instead')
```
lukasIO, line 113: we'll want some more info here in the deprecation notice, pointing to a guide (that I'll write today) that outlines the differences

lib/src/core/engine.dart
```diff
@@ -636,15 +736,40 @@ class Engine extends Disposable with EventsEmittable<EngineEvent> {
...
+      if (decryptedData == null) {
+        logger.warning('Failed to decrypt data packet');
+        return;
+      }
+
+      final decryptedPacketPayload =
+          lk_models.EncryptedPacketPayload.fromBuffer(decryptedData);
+      final newDp = asDataPacket(decryptedPacketPayload);
+
+      _emitDataPacket(newDp);
```
lukasIO, line 766: we should add some encryption metadata here, I opted to add the `encryptionType` for the data packet. 

This allows us also to error on data streams that have inconsistent encryption types during the lifetime of a stream

PR #891, opened 2025-10-14
move the accessToken to header

xianshijing-lk:
Move the access_token to header in the format of 'Authorization': 'Bearer $token'

### DIFF
--- a/lib/src/core/room.dart
+++ b/lib/src/core/room.dart
@@ -466,7 +466,7 @@ class Room extends DisposableChangeNotifier with EventsEmittable<RoomEvent> {
       for (final info in event.response.otherParticipants) {
         logger.fine('Creating RemoteParticipant: sid = ${info.sid}(identity:${info.identity}) '
             'tracks:${info.tracks.map((e) => e.sid)}');
-        _getOrCreateRemoteParticipant(info.identity, info);
+        await _getOrCreateRemoteParticipant(info.identity, info);
       }
 
       if (e2eeManager != null && event.response.sifTrailer.isNotEmpty) {
--- a/lib/src/options.dart
+++ b/lib/src/options.dart
@@ -152,6 +152,7 @@ class RoomOptions {
     bool? dynacast,
     bool? stopLocalTrackOnUnpublish,
     E2EEOptions? e2eeOptions,
+    E2EEOptions? encryption,
     bool? fastPublish,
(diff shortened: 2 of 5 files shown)


xianshijing-lk:
Thanks @hiroshihorie, do we have any setups for the flutter format ? like using a script or some setups on Android Studio ?

xianshijing-lk:
flutter analyze gives me the same errors on main, I put in some lines of code in analysis_options.yaml to suppress the warnings instead.

xianshijing-lk:
Hi @hiroshihorie  @cloudwebrtc , could you please help review this PR?

hiroshihorie:
@xianshijing-lk You just need to run `dart format .`

example/analysis_options.yaml
```diff
@@ -31,6 +31,11 @@ analyzer:
     constant_identifier_names: ignore
     no_wildcard_variable_uses: ignore
     use_super_parameters: ignore
+    avoid_print: ignore
+    use_build_context_synchronously: ignore
+    discarded_futures: ignore
```
hiroshihorie: `discarded_futures` are something we should fix soon for safer code, but it's fine to suppress it temporary if we don't forget about it.

xianshijing-lk: There are quite a few instance of discarded_futures warnings, let me take a follow-up to address it and re-enable the warnings. 

Added a TODO here.

xianshijing-lk:
I rebased then got 4 more analyze warnings, I just fixed them as I think some of them worth fixing now
   info • Missing an 'await' for the 'Future' computed by this expression • lib/src/core/room.dart:469:9 • unawaited_futures
   info • Unnecessary 'this.' qualifier • lib/src/options.dart:168:33 • unnecessary_this
   info • An uninitialized variable should have an explicit type annotation • web/e2ee.frame_cryptor.dart:179:9 • prefer_typing_uninitialized_variables
   info • Use a function declaration rather than a variable assignment to bind a function to a name • web/e2ee.worker.dart:118:9 • prefer_function_declarations_over_variables
   
   
   I think the next step might be address all those issues and re-enable the analyze warnings.

hiroshihorie:
@xianshijing-lk LGTM, yes I'll fix other analyze errors in another branch. The test not passing is due to recent changes.

PR #896, opened 2025-10-18
fix: fix async call for update participant info

cloudwebrtc:
close https://github.com/livekit/client-sdk-flutter/issues/895


cloudwebrtc:
created on the wrong base branch, close and recreate

PR #897, opened 2025-10-18
fix: fix async call for update participant info

cloudwebrtc:
close https://github.com/livekit/client-sdk-flutter/issues/895


hiroshihorie:
looks like `TrackPublishedEvent` gets emitted before `ParticipantConnectedEvent` and e2e test is failing now, will make a fix.

PR #899, opened 2025-10-20
Enable to translate labels for screen select dialog.

lawnbot:
To build a multilingual app, developers should be able to translate the labels in the ScreenSelectDialog.


hiroshihorie (approved):
✅

PR #908, opened 2025-11-03
fix: `CancelListenFunc` should have a declared return type

spydon:
`CancelListenFunc` should have a declared return type, otherwise `dynamic` will start trickling into the users' of this code bases.


hiroshihorie (approved):
✅ Nice find thanks

PR #916, opened 2025-11-11
Refactor enums

hiroshihorie:


hiroshihorie:
This breaks a lot of symbols so, will close this for now. Perhaps better when a major update.

PR #929, opened 2025-11-24
chore: Bump dependencies

spydon:
This PR sets the Flutter min version to the correct version that Dart 3.6.0 was used with, so previous Flutter versions can't be used anyways.

It also bumps major versions of `connectivity_plus` and `device_info_plus` so that users of this package aren't stuck on older versions.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -20,7 +20,7 @@ homepage: https://github.com/livekit/client-sdk-flutter
 
 environment:
   sdk: ">=3.6.0 <4.0.0"
-  flutter: ">=3.27.1"
+  flutter: ">=3.27.0"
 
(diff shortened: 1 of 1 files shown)


pubspec.yaml
```diff
@@ -20,7 +20,7 @@ homepage: https://github.com/livekit/client-sdk-flutter
 
 environment:
   sdk: ">=3.6.0 <4.0.0"
-  flutter: ">=1.17.0"
+  flutter: ">=3.27.1"
```
spydon: This is the min flutter version corresponding to SDK >=3.6

hiroshihorie: I think it's `3.27.0` 🙂

PR #936, opened 2025-12-02
v2.6.0

hiroshihorie:


rokk4:
Would following Semantic Versioning not mean this would be 2.6.0 since session api is a new feature?

Issue #962, opened 2026-01-20
Incorrect and no-op microphone mute state on call reconnection


coder-with-a-bushido:
**Describe the bug**
Incorrect mute state and unable to change mute state after call reconnection if mic was unmuted and then muted once before network outage. But it works fine if mic was unmuted before network outage.

**To Reproduce**
1. Join call
2. Unmute mic once to get the audio stream started
3. Mute mic now
4. Turn off internet for 5+ seconds
5. Now see the mic state changing automatically to unmuted (incorrectly)
6. Try to mute it, it won't work

**Expected behavior**
Mute state shouldn't change after network reconnection and mute/unmute should work too

**Platform information**

* **Flutter version**: 
 ```
[✓] Flutter (Channel stable, 3.38.7, on macOS 26.2 25C56 darwin-arm64, locale
    en-IN) [345ms]
    • Flutter version 3.38.7 on channel stable at
      /Users/mymac/development/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 3b62efc2a3 (7 days ago), 2026-01-13 13:47:42 -0800
    • Engine revision 78fc3012e4
    • Dart version 3.10.7
    • DevTools version 2.51.1
    • Feature flags: enable-web, enable-linux-desktop, no-enable-macos-desktop,
      enable-windows-desktop, enable-android, enable-ios, cli-animations,
      enable-native-assets, omit-legacy-version-file, enable-lldb-debugging

[✓] Android toolchain - develop for Android devices (Android SDK version 35.0.0)
    [1,446ms]
    • Android SDK at /Users/mymac/Library/Android/sdk
    • Emulator version 36.2.11.0 (build_id 14212360) (CL:N/A)
    • Platform android-36, build-tools 35.0.0
    • Java binary at: /Applications/Android
      Studio.app/Contents/jbr/Contents/Home/bin/java
      This is the JDK bundled with the latest Android Studio installation on
      this machine.
      To manually set the JDK path, use: `flutter config
      --jdk-dir="path/to/jdk"`.
    • Java version OpenJDK Runtime Environment (build 21.0.8+-14018985-b1038.68)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 26.1.1) [742ms]
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 17B100
    • CocoaPods version 1.16.2

[✓] Chrome - develop for the web [9ms]
[...]

PR #963, opened 2026-01-20
fix: set proper mute state for audio/video tracks after reconnection

coder-with-a-bushido:
Closes https://github.com/livekit/client-sdk-flutter/issues/962

### How does it fix that?

The methods `publishAudioTrack` and `publishVideoTrack` are called on call reconnection here -
https://github.com/livekit/client-sdk-flutter/blob/70d4d1f5fdc6f565cbe565e3d5a50d5c23b5ddef/lib/src/participant/local.dart#L560-L570

But inside the methods we're not passing the `muted` state from the old tracks we have. It defaults to `false` because it's not set in protobuf:
https://github.com/livekit/client-sdk-flutter/blob/70d4d1f5fdc6f565cbe565e3d5a50d5c23b5ddef/lib/src/proto/livekit_rtc.pb.dart#L1153

By passing it correctly, this PR fixes the bug in the linked issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where the muted state of audio and video tracks was not being correctly propagated when publishing to the server.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/lib/src/participant/local.dart
+++ b/lib/src/participant/local.dart
@@ -134,6 +134,7 @@ class LocalParticipant extends Participant<LocalTrackPublication> {
       name: publishOptions.name ?? AudioPublishOptions.defaultMicrophoneName,
       type: track.kind.toPBType(),
       source: track.source.toPBType(),
+      muted: track.muted,
       stream: buildStreamId(publishOptions, track.source),
       disableDtx: !publishOptions.dtx,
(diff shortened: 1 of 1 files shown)


hiroshihorie:
Nice!, thank you for your PR, I'll test this out.

PR #970, opened 2026-01-27
add control mic option to SessionOptions

NTJ3:
## Summary by CodeRabbit

* **New Features**
  * Sessions now expose a configurable options property that can be updated at runtime.
  * Added a new session option to control whether the microphone is enabled by default (defaults to enabled).

* **Behavior Change**
  * Microphone will only be auto-enabled on startup when the session's default-microphone option is true; updating options triggers immediate state updates.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/.changes/ci-ios-platform-download
+++ b/.changes/ci-ios-platform-download
@@ -1,0 +1,1 @@
+patch type="chore" "Skip iOS platform download in CI"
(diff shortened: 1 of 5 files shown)


NTJ3:
Draft

PR #1057, opened 2026-04-17
fix(android): prevent screen from turning off during active video calls in example application

mechtech-mind:
This PR fixes an issue in the Android example application where the device screen turns off automatically during an active video call due to screen timeout.

### **Problem**
During an ongoing LiveKit call on Android, the screen is not kept awake, which leads to:

- Poor user experience during video calls
- Interrupted visibility when no user interaction occurs

### **Solution**
This PR adds native Android window flag handling in MainActivity to ensure the screen remains active while a call is in progress:

- Uses native flags to prevent screen timeout
- Uses native flags as well to handle lock screen visibility

### **The flags are:**
- **Enabled** when the room is active
- **Cleared** when the room is disposed


mechtech-mind (commented):
@hiroshihorie is it ok?

mechtech-mind (commented):
yeah looks ok

example/lib/pages/room.dart
```diff
@@ -34,6 +35,7 @@ class _RoomPageState extends State<RoomPage> {
   @override
   void initState() {
     super.initState();
+    unawaited(enableInCallFlags());
```
hiroshihorie: This channel only exists on Android, so this will throw on iOS/web/desktop. Could we wrap both enable/disable calls in the existing Android platform check?

example/android/app/src/main/kotlin/io/livekit/example/MainActivity.kt
```diff
@@ -1,6 +1,68 @@
...
+    }
+
+    private fun setInCallScreenFlags() {
+        val window = window
+
+        // Keep screen on during call
+        window.addFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON)
+
+        // Show over lock screen (modern way)
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O_MR1) {
```
hiroshihorie: For the timeout fix, I think `FLAG_KEEP_SCREEN_ON` is the important part. Showing/waking over the lock screen feels like a separate behavior change, so I'd leave that out unless we really want it.

pubspec.lock
```diff
@@ -436,10 +436,10 @@ packages:
     dependency: transitive
     description:
       name: matcher
-      sha256: dc0b7dc7651697ea4ff3e69ef44b0407ea32c487a39fff6a4004fa585e901861
+      sha256: "12956d0ad8390bbcc63ca2e1469c0619946ccb52809807067a7020d57e647aa6"
       url: "https://pub.dev"
     source: hosted
-    version: "0.12.19"
+    version: "0.12.18"
```
hiroshihorie: This lockfile change looks unrelated to the example keep-awake fix. Can we drop it from the PR?

hiroshihorie (commented):
Left a few small notes. The keep-awake idea makes sense; I think this just needs a little tightening before merge.

example/lib/pages/room.dart
mechtech-mind: yes thanks for the insight. I will wrap it up in platform check for android only.

example/android/app/src/main/kotlin/io/livekit/example/MainActivity.kt
mechtech-mind: yes that is also correct. I implemented it as I was thinking it from the incoming call push notification perspective when app is in the terminated state.

We can definitely drop it here. i will amke sure to make the necessary changes.

pubspec.lock
mechtech-mind: yes I will drop it as well.

mechtech-mind:
Hi @hiroshihorie, 

I have updated the PR as per the changes you suggested:

wrapped the in-call flag calls with Android platform checks
removed the lockscreen/wake-over-screen behavior
removed the unrelated lockfile changes

Thanks again for the review.

hiroshihorie (commented):
Thanks for the updates. The lockfile and lock screen parts look fixed. I left two small platform notes.

example/lib/pages/room.dart
```diff
@@ -1,8 +1,10 @@
 import 'dart:async';
 import 'dart:convert';
+import 'dart:io';
```
hiroshihorie: This adds `dart:io`, which will break the web example. Could we use `lkPlatformIs(PlatformType.android)` here instead?

example/lib/pages/room.dart
```diff
@@ -63,6 +70,7 @@ class _RoomPageState extends State<RoomPage> {
     widget.room.removeListener(_onRoomDidUpdate);
     unawaited(_disposeRoomAsync());
     onWindowShouldClose = null;
+    unawaited(disableInCallFlags());
```
hiroshihorie: Small follow-up: this still runs on every platform. Can we guard the disable call with the same Android check as enable?

mechtech-mind: Thanks again, I forgot to guard the disable flag.  I will update accordingly.

example/lib/pages/room.dart
mechtech-mind: okay I did not know about this. I will update it.

mechtech-mind:
@hiroshihorie Thanks again for the reviews. I have updated the file as per the requirements. I will make sure to check thoroughly before creating the PR.

hiroshihorie (approved):

PR #1082, opened 2026-05-12
chore: Update deps

hiroshihorie:


cloudwebrtc (approved):

hiroshihorie:
Closing as stale. The version bumps here are long superseded (flutter_webrtc is pinned at 1.6.0 on main, dart_webrtc 1.8.1 landed via #1163) and the constraint ranges no longer match what main carries. If widening floors for downstream consumers becomes needed it should be redone fresh against current constraints.

PR #1085, opened 2026-05-15
Add RPC tester interface to example app

1egoman:
Adds a new button on the bottom control bar for the "rpc tester":

Screenshot 2026-05-15 at 11 09 19 AM

When clicked, it opens up a new panel which has two sections:

Screenshot 2026-05-15 at 11 09 04 AM

The top section allows sending RPC messages - select a destination participant, topic, and enter a payload, and click "Send". Note that there are two payload presets - `"hello world"` and `"X" * 20000` to provide some easy to generate data to test some edge cases.

The bottom section lets you configure rpc handlers. Enter a rpc handler topic and a static response, and submit. Then when this participant receives a RPC request, the request will be responded to with the static response. This static response also has the same payload presets as the request.

> [!WARNING]
> This pull request was LLM generated and has only been lightly reviewed by a human. The author has tested this and confirms it works in the happy path, but no other validation has been done.
> 
> A more thorough review of this needs to occur before it could be merged.


example/lib/widgets/rpc_test_sheet.dart
```diff
@@ -0,0 +1,649 @@
...
+          crossAxisAlignment: CrossAxisAlignment.start,
+          children: [
+            Row(
+              children: [
+                Expanded(
+                  child: Text(
+                    entry.topic,
+                    style: theme.textTheme.titleSmall?.copyWith(fontWeight: FontWeight.bold, color: theme.scaffoldBackgroundColor),
+                  ),
+                ),
```
1egoman, line 552: A drive by comment: the colors in this app (besides being out of date) are fairly limited, which makes styling some of these more complicated interfaces fairly challenging.

I think I might make a follow up pull request which updates the example to use the more modern color pallete / logo, and at the same time try to add in some more color scales to make this less annoying.

1egoman: I did this: https://github.com/livekit/client-sdk-flutter/pull/1086

hiroshihorie:
Updated this branch to bring it back to life. The tester itself already landed on main through #1087, which included a hardened copy of this file, so the merge resolves to main's version and keeps the original commits in history. On top of that this PR now carries follow-up fixes to the tester: registration is ordered so a method owned by another component can't leave a phantom card, the per handler invocation log is capped, in flight handlers no longer notify after disposal, and the destination dropdown stays valid when the selected participant leaves.

hiroshihorie (approved):
Reviewed the follow-up fixes, all good. The original tester landed via #1087.

PR #1087, opened 2026-05-15
Add support for RPC V2

1egoman:
See this pull request for more info about what this does and how rpc v2 works: https://github.com/livekit/client-sdk-js/pull/1832

Also, for an in depth description of behavior in edge cases, see [RPC_SPEC.md here](https://github.com/livekit/client-sdk-js/blob/00e93de0c3d85f846fa6707749fa46457a653a29/RPC_SPEC.md).

> [!WARNING]
> This pull request was LLM generated and has only been lightly reviewed by the author, who is not a flutter expert. I have tested this and can confirm it works in the happy path, but no other validation has been done.
> 
> A more thorough review of this needs to occur before it could be merged.


lib/src/rpc/rpc_client_manager.dart
```diff
@@ -0,0 +1,274 @@
...
+    final effectiveTimeout = Duration(
+      milliseconds: (params.responseTimeoutMs.inMilliseconds - maxRoundTripLatency.inMilliseconds)
+          .clamp(minEffectiveTimeout.inMilliseconds, double.infinity)
+          .toInt(),
+    );
+
+    final destination = params.destinationIdentity;
+    final remoteParticipant = _room.getParticipantByIdentity(destination);
+    final remoteProtocol = remoteParticipant?.clientProtocol ?? ClientProtocolVersion.v0.toIntValue();
+    final useV2 = remoteProtocol >= ClientProtocolVersion.v1.toIntValue();
```
1egoman: thought: I wonder if it would be cleaner to make `ClientProtocolVersion` implement `Comparable` ([this](https://api.dart.dev/dart-core/Comparable-class.html)) to avoid all the `toIntValue()`s here?

test/core/rpc_test.dart
1egoman, line 36: question: Heads up that I kept all the old / pre-existing rpc v1 tests around. I'm not sure though if this is really still needed, because the new tests should be testing all the same paths. I'll defer to what others think here, I don't have a strong opinion on if they should be kept or removed.

1egoman:
^ In the above "breaking change" warning, it looks like the issue is:

```
BREAKING CHANGES
└─┬ Class Participant
  └── Field "clientProtocol" added (required) (CF02)
```

But it's a field addition... That shouldn't be a breaking change? Let me know what I am missing - I could I think make this private, but I'd need to verify that is ok first.

example/lib/widgets/rpc_test_sheet.dart
```diff
@@ -0,0 +1,650 @@
...
+        child: Column(
+          crossAxisAlignment: CrossAxisAlignment.start,
+          children: [
+            Row(
+              children: [
+                Expanded(
+                  child: Text(
+                    entry.topic,
+                    style: theme.textTheme.titleSmall
+                        ?.copyWith(fontWeight: FontWeight.bold, color: theme.scaffoldBackgroundColor),
```
xianshijing-lk: nit, why using scaffoldBackgroundColor as the text color ?

example/lib/widgets/rpc_test_sheet.dart
```diff
@@ -0,0 +1,650 @@
...
+              controller: _responseCtl,
+              minLines: 1,
+              maxLines: 4,
+              decoration: const InputDecoration(
+                labelText: 'Static response',
+                border: OutlineInputBorder(),
+                isDense: true,
+              ),
+              onChanged: (value) => widget.controller.updateStaticResponse(entry.topic, value),
+              style: TextStyle(color: Theme.of(context).scaffoldBackgroundColor),
```
xianshijing-lk: ditto

example/lib/widgets/rpc_test_sheet.dart
```diff
@@ -0,0 +1,650 @@
...
+
+  List<RpcHandlerEntry> get handlers => List.unmodifiable(_handlers);
+
+  bool isRegistered(String topic) => _handlers.any((h) => h.topic == topic);
+
+  void registerHandler(Room room, String topic, String staticResponse) {
+    if (isRegistered(topic)) {
+      return;
+    }
+    _room = room;
```
xianshijing-lk: question, why do we need to assign a new room to _room for each registerHandler ?

example/lib/widgets/rpc_test_sheet.dart
```diff
@@ -0,0 +1,650 @@
...
+    final entry = RpcHandlerEntry(topic: topic, staticResponse: staticResponse);
+    _handlers.add(entry);
+    room.registerRpcMethod(topic, (data) async {
```
xianshijing-lk, line 64: should we still call room.registerRpcMethod() if a room is town down already ?

lib/src/core/room.dart
```diff
@@ -122,8 +124,10 @@ class Room extends DisposableChangeNotifier with EventsEmittable<RoomEvent> {
   // Agents
   final Map<String, DateTime> _transcriptionReceivedTimes = {};
 
-  // RPC Handlers
-  final Map<String, RpcRequestHandler> _rpcHandlers = {};
+  // RPC managers (caller and handler roles)
+  @internal
+  late final RpcClientManager rpcClientManager;
```
xianshijing-lk: should it be _rpcClientManager? 

I think it is missing _

lib/src/core/room.dart
```diff
@@ -139,12 +143,19 @@ class Room extends DisposableChangeNotifier with EventsEmittable<RoomEvent> {
   // Pending subscriber tracks keyed by participantSid, for tracks arriving before metadata or before the room connected.
   late final PendingTrackQueue _pendingTrackQueue;
 
-  // for testing
+  // for testing — pass-through to the server manager's handler map.
   @internal
-  Map<String, RpcRequestHandler> get rpcHandlers => _rpcHandlers;
+  Map<String, RpcRequestHandler> get rpcHandlers => _rpcServerManager.handlers;
```
xianshijing-lk: nit, are callers supposed to be able to mutate it ? 

If not, it is slightly better to do 
Map<String, RpcRequestHandler> get rpcHandlers => Map.unmodifiable(_rpcServerManager.handlers);

lib/src/participant/participant.dart
```diff
@@ -97,6 +97,12 @@ abstract class Participant<T extends TrackPublication> extends DisposableChangeN
     return DateTime.timestamp();
   }
 
+  /// Client-to-client protocol version this participant supports, as advertised in
+  /// `ParticipantInfo.clientProtocol`. Drives the caller- and handler-side decision
+  /// between RPC v1 packets and RPC v2 data streams. Absent / unrecognized values
+  /// from older clients resolve to `0` (matches [ClientProtocolVersion.v0]).
+  int get clientProtocol => _participantInfo?.clientProtocol ?? ClientProtocolVersion.v0.toIntValue();
```
xianshijing-lk: Just want to make sure I understand, this _participantInfo is the remote participant info, right ?

lib/src/rpc/rpc_client_manager.dart
```diff
@@ -0,0 +1,271 @@
...
+  /// destination participant's advertised `clientProtocol`.
+  Future<String> performRpc(PerformRpcParams params) async {
+    final requestId = _uuid.v4();
+    final completer = Completer<String>();
+
+    const maxRoundTripLatency = Duration(seconds: 7);
+    const minEffectiveTimeout = Duration(milliseconds: 1000);
+    final effectiveTimeout = Duration(
+      milliseconds: (params.responseTimeoutMs.inMilliseconds - maxRoundTripLatency.inMilliseconds)
+          .clamp(minEffectiveTimeout.inMilliseconds, double.infinity)
```
xianshijing-lk: nit, this seems to mix int and double, maybe it is better to do 
final effectiveMs = math.max(
  minEffectiveTimeout.inMilliseconds,
  params.responseTimeoutMs.inMilliseconds - maxRoundTripLatency.inMilliseconds,
);
final effectiveTimeout = Duration(milliseconds: effectiveMs);

lib/src/rpc/rpc_client_manager.dart
xianshijing-lk: Good idea, lets do it

lib/src/rpc/rpc_client_manager.dart
```diff
@@ -0,0 +1,271 @@
...
+    required String method,
+    required String payload,
+    required Duration responseTimeout,
+  }) async {
+    final local = _room.localParticipant;
+    if (local == null) {
+      throw RpcError(code: RpcError.sendFailed, message: 'No local participant');
+    }
+
+    final writer = await local.streamText(StreamTextOptions(
```
xianshijing-lk, line 163: can you tell me where this streamText() is implemented ? local_participant.dart ?

xianshijing-lk (commented):
some nits / questions

hiroshihorie (approved):

PR #1096, opened 2026-05-25
Default VideoTrackRenderer renderMode to auto

hiroshihorie:
Changes the default `renderMode` from `texture` to `auto`. For now `auto` resolves to texture rendering on all platforms, so behavior is unchanged. This lets the SDK change how `auto` resolves in a future release without an API break.

Platform view support and related fixes from the original diff already landed via #1116 and #1135.


lib/src/widgets/video_track_renderer.dart
```diff
@@ -178,12 +181,23 @@ class _VideoTrackRendererState extends State<VideoTrackRenderer> {
...
+  double? get _rendererAspectRatio {
+    final renderer = _renderer;
+    if (renderer is rtc.RTCVideoRenderer) {
+      return renderer.videoValue.aspectRatio;
+    }
+    if (renderer is rtc.RTCVideoPlatformViewController) {
+      return renderer.value.aspectRatio;
+    }
+    return null;
+  }
```
bdlukaa: This fixes this bug:

```bash
TypeError: type 'RTCVideoPlatformViewController' is not a subtype of type 'RTCVideoRenderer?' in type cast
#0      _VideoTrackRendererState._attach.<fn>.<fn> (package:livekit_client/src/widgets/video_track_renderer.dart:192)
#1      State.setState (package:flutter/src/widgets/framework.dart:1199)
#2      _VideoTrackRendererState._attach.<fn> (package:livekit_client/src/widgets/video_track_renderer.dart:191)
#3      RTCVideoPlatformViewController.eventListener (package:flutter_webrtc/src/native/rtc_video_platform_view_controller.dart:133)
#4      _RootZone.runUnaryGuarded (dart:async/zone.dart:891)
#5      _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:381)
#6      _DelayedData.perform (dart:async/stream_impl.dart:573)
#7      _PendingEvents.handleNext (dart:async/stream_impl.dart:678)
#8      _PendingEvents.schedule.<fn> (dart:async/stream_impl.dart:649)
#9      _microtaskLoop (dart:async/schedule_microtask.dart:40)
#10      _startMicrotaskLoop (dart:async/schedule_microtask.dart:49)
```

hiroshihorie: This fix landed on main via #1116 and #1135, and this branch now includes it after merging main. Thanks for flagging it.

PR #1098, opened 2026-05-29
Fix pixel density in AdaptiveStream

hiroshihorie:
The adaptive-stream visibility observer fed each view's logical (DIP) size straight to the server, so on retina/HiDPI displays it under-requested by the device pixel ratio and the server returned an upscaled/soft layer. Server layers are sized in physical pixels.

- Add `AdaptiveStreamPixelDensity` (`auto` | `fixed(double)`) mirroring the JS SDK's `pixelDensity` option, plus `VideoTrackRenderer.adaptiveStreamPixelDensity` (default: `auto`), set per view rather than room-wide. Fractional densities are supported and the resolved value is capped at 3x to bound bandwidth.
- Track each rendered view with a `VideoTrackViewRegistration` that owns both the `GlobalKey` and pixel density, so density updates stay tied to the renderer lifecycle.
- Resolve auto mode from each view's `devicePixelRatio` via `MediaQuery.maybeDevicePixelRatioOf`; fixed modes use a constant multiplier.
- `_computeVideoViewVisibility` scales each view's logical size by the resolved density, materializes the visible sizes once per tick, and requests the largest scaled dimensions across all views.


lib/src/track/local/local.dart
```diff
@@ -40,22 +40,33 @@ import 'video.dart';
...
+  }) {
     final key = GlobalKey();
-    viewKeys.add(key);
+    viewPixelDensities[key] = pixelDensity;
     return key;
   }
 
+  @internal
+  void updateViewKeyPixelDensity(GlobalKey key, AdaptiveStreamPixelDensity pixelDensity) {
+    if (viewPixelDensities.containsKey(key)) viewPixelDensities[key] = pixelDensity;
```
xianshijing-lk: should we log or handle the case where if (!viewPixelDensities.containsKey(key)) ?

xianshijing-lk (changes requested):

xianshijing-lk (approved):
one question, lgtm in general

lib/src/track/local/local.dart
hiroshihorie: Good point 🙂 I changed this to keep the key and pixel density together in VideoTrackViewRegistration, so there is no separate map lookup anymore. Missing context/size is still skipped as part of the normal widget lifecycle.

cloudwebrtc (approved):

PR #1120, opened 2026-06-29
fix: improve initial video quality by setting x-google-start-bitrate for all video codecs

xianshijing-lk:
## Summary
- Apply `x-google-start-bitrate` SDP hint to all video codecs (VP8, VP9, AV1, H264, H265), not just SVC codecs
- Use 90% of target bitrate as start bitrate to prevent initial blurriness
- Degradation preference already defaults to `maintainResolution`

## Problem
Video starts blurry for 5-15 seconds before improving. This is caused by WebRTC's bandwidth estimator starting at ~300kbps and slowly ramping up to the target bitrate.

## Solution
**x-google-start-bitrate**: Tell WebRTC to start at 90% of target bitrate instead of ramping up from ~300kbps. Applied consistently to all video codecs.

 Note, The PR only addresses the x-google-start-bitrate fix (changing from SVC-only to all video codecs, and using 0.9 multiplier)
  - The maintainResolution default was already in place in Flutter SDK, so no change was needed for that part
  

## Test plan
- [ ] Verify video quality is sharp from the start when publishing
- [ ] Test with VP8, VP9, H264, AV1 codecs
- [ ] Verify bandwidth estimator adapts properly if network can't handle the start bitrate

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/src/core/transport.dart
+++ b/lib/src/core/transport.dart
@@ -29,13 +29,17 @@ import '../utils.dart';
 
 const ddExtensionURI = 'https://aomediacodec.github.io/av1-rtp-spec/#dependency-descriptor-rtp-header-extension';
 
-/* The svc codec (av1/vp9) would use a very low bitrate at the begining and
-increase slowly by the bandwidth estimator until it reach the target bitrate. The
-process commonly cost more than 10 seconds cause subscriber will get blur video at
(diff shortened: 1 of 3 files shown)


cloudwebrtc (approved):
lgtm

PR #1121, opened 2026-07-02
fix: allow selectAudioOutput on Android

yash-garg:
`Helper.selectAudioOutput` already routes audio on Android via device ID strings (speaker, earpiece, bluetooth, wired-headset). The desktop-only gate was silently swallowing the call and leaving `room.setAudioOutputDevice()` as a no-op on Android.

### DIFF
--- a/.changes/fix-android-select-audio-output
+++ b/.changes/fix-android-select-audio-output
@@ -1,0 +1,1 @@
+patch type="fixed" "Allow selectAudioOutput on Android"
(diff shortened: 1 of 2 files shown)


hiroshihorie (approved):
Thanks for the fix! Verified the claim against flutter-webrtc 1.5.2, the version this SDK pins. Android's enumerateDevices reports audiooutput deviceIds that are exactly the AudioDeviceKind type names (speaker, earpiece, bluetooth, wired-headset), and the selectAudioOutput handler maps them back through AudioDeviceKind.fromTypeName into AudioSwitchManager, so the round trip works. Unknown ids are a safe no-op. Web keeps using setSinkId in room.dart and iOS routing stays with AudioManager, so warning on those two platforms is correct.

Pushed a small commit adding the .changes entry per repo convention.

PR #1141, opened 2026-07-28
fix: don't set zero size if size is unknown

holzgeist:
Sending zero instead of null confuses other sdks (at least rust), e.g. [here](https://github.com/livekit/rust-sdks/blob/e37adc1dc4cb90322733fcb0fe1155262f309ff1/livekit-data-stream/src/incoming.rs#L325)

### DIFF
--- a/.changes/fix-data-stream-size
+++ b/.changes/fix-data-stream-size
@@ -1,0 +1,1 @@
+patch type="fixed" "Omit data stream totalLength when size is unknown"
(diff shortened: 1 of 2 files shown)


hiroshihorie:
Pushed two small additions: streamBytes had the same issue so it now omits totalLength the same way, and added a changeset for the CI check. Verified against the generated protobuf code that null skips the field and that the old code did put an explicit 0 on the wire, JS and Swift and Rust all omit it when unknown. Thanks for the fix!

hiroshihorie (approved):

PR #1149, opened 2026-08-03
chore: introduce DevelopmentTokenSource and deprecate SandboxTokenSource

hiroshihorie:
## What

Port of livekit/client-sdk-swift#1077. LiveKit Cloud renamed the sandbox token server to the development token server, this aligns the API:

- `DevelopmentTokenSource({required String id})` is the new name, same endpoint and header
- `SandboxTokenSource` remains as a deprecated subclass. Unlike Swift, a typealias is not enough here because the old constructor takes `sandboxId:` while the new one takes `id:`, so the subclass keeps existing call sites compiling unchanged
- Doc references updated to the development token server wording

Also excludes `build/` from analysis, Xcode vendors Swift package checkouts there when SPM is enabled (same fix already applied to components-flutter and the agent starter).

## Notes

- Changeset: `minor type="added"`
- The JS SDK has the same concept (`TokenSource.sandboxTokenServer`) and has not been renamed yet, tracked separately

### DIFF
--- a/lib/livekit_client.dart
+++ b/lib/livekit_client.dart
@@ -88,7 +88,7 @@ export 'src/token_source/literal.dart';
 export 'src/token_source/endpoint.dart';
 export 'src/token_source/custom.dart';
 export 'src/token_source/caching.dart';
-export 'src/token_source/sandbox.dart';
+export 'src/token_source/development.dart';
 export 'src/token_source/jwt.dart';
 
 /// Misspelled alias for [kRpcVersion]. Kept for backwards compatibility with code
--- a/lib/src/token_source/development.dart
+++ b/lib/src/token_source/development.dart
@@ -17,7 +17,7 @@ import 'endpoint.dart';
 /// A token source that queries LiveKit's development token server for development and testing.
 ///
 /// This token source connects to LiveKit Cloud's
-/// [development token server](https://cloud.livekit.io/projects/p_/sandbox/templates/token-server),
+/// [development token server](https://docs.livekit.io/frontends/build/authentication/sandbox-token-server/),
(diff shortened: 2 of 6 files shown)


lib/src/token_source/development.dart
```diff
@@ -14,30 +14,42 @@
 
 import 'endpoint.dart';
 
-/// A token source that queries LiveKit's sandbox token server for development and testing.
+/// A token source that queries LiveKit's development token server for development and testing.
 ///
-/// This token source connects to LiveKit Cloud's sandbox environment, which is perfect for
-/// quick prototyping and getting started with LiveKit development.
+/// This token source connects to LiveKit Cloud's
+/// [development token server](https://cloud.livekit.io/projects/p_/sandbox/templates/token-server),
```
lukasIO: let's use the right docs link for this 


```suggestion
/// [development token server](https://docs.livekit.io/frontends/build/authentication/sandbox-token-server/),
```

lib/livekit_client.dart
```diff
@@ -88,7 +88,7 @@ export 'src/token_source/literal.dart';
 export 'src/token_source/endpoint.dart';
 export 'src/token_source/custom.dart';
 export 'src/token_source/caching.dart';
-export 'src/token_source/sandbox.dart';
```
MaxHeimbrock, line 91: Will this break clients using the old path?

hiroshihorie: It's only renamed, and should be ok.

MaxHeimbrock (approved):
I tested it in the flutter agent starter and it looks good. Deprecation works, code is still running but IDE shows the deprecation.

PR #1152, opened 2026-08-05
fix: Room.getSid() never resolves when the JoinResponse carries an empty room sid

### LINKED ISSUE #1151: Room.getSid() never resolves when the JoinResponse carries an empty room sid — listens for SignalRoomUpdateEvent on the wrong emitter (deadlocks audio publish via the Krisp noise filter)
## Summary

`Room.getSid()` registers its `SignalRoomUpdateEvent` listener on the **Room's** event emitter, but that event is only ever emitted on the **signal client's** emitter — nothing forwards it. So whenever `getSid()` is called while the room sid is still empty, the returned future **never completes** (until room disconnect, which completes it with `''`).

The sid arriving later via RoomUpdate updates `_roomInfo` (so a *fresh* `getSid()` call returns it), but it never resolves the already-returned future.

Code (permalinks to current `main`, 9e05b79):

- `getSid()` listens on `events` (the Room emitter, typed for `RoomEvent`s): https://github.com/livekit/client-sdk-flutter/blob/9e05b7987f26de415750b3c12d0e4c889a64f920/lib/src/core/room.dart#L1106-L1110
- `SignalRoomUpdateEvent` is emitted only on the signal client's emitter: https://github.com/livekit/client-sdk-flutter/blob/9e05b7987f26de415750b3c12d0e4c889a64f920/lib/src/core/signal_client.dart#L317
- The Room consumes it via `_signalListener`, not via `events`: https://github.com/livekit/client-sdk-flutter/blob/9e05b7987f26de415750b3c12d0e4c889a64f920/lib/src/core/room.dart#L434

## When does this bite?

Room sids are assigned asynchronously by the server: the JoinResponse can carry `room.sid == ""`, with the real sid following in a RoomUpdate (this is the reason `Room.sid` became `await room.getSid()` in v2.0.0). Any `await room.getSid()` during that post-join window deadlocks.

## Real-world impact (how we found it)

`livekit_noise_filter`'s `onPublish` awaits `room.getSid()`, and `_publishAudioTrack` awaits the processor **after** addTrack + negotiation + `addTrackPublication`, inside the `SerialRunner` publish lock:

- processor await: https://github.com/livekit/client-sdk-flutter/blob/9e05b7987f26de415750b3c12d0e4c889a64f920/lib/src/participant/local.dart#L239
- publish lock: https://github.com/livekit/client-sdk-flutter/blob/9e05b7987f26de415750b3c12d0e4c889a64f920/lib/src/participant/local.dart#L68 with https://github.com/livekit/client-sdk-flutter/blob/9e05b7987f26de415750b3c12d0e4c889a64f920/lib/src/support/serial_runner.dart#L38-L40

Consequences observed in production field logs (macOS, livekit_client 2.8.1 + livekit_noise_filter 0.2.1, LiveKit Cloud, mic published ~90ms after connect):

1. The microphone track publishes, is unmuted, and audio flows to remote participants.
2. `setMicrophoneEnabled(true)` never returns.
3. The hung future holds `_publishRunner` **forever**, so every subsequent `setMicrophoneEnabled` / `setCameraEnabled` / `setSourceEnabled` call also hangs — **the user cannot mute their live microphone for the rest of the session**. For us (telehealth) that's a privacy incident, not a cosmetic bug.

It's race-dependent (publish must reach `getSid()` before the sid-bearing RoomUpdate lands), which is probably why it hasn't been reported: the audio *works*, so the symptom looks like a stuck spinner or a dead mute button and gets blamed on the network.

## Reproduction

Deterministic repro using the SDK's own mock harness (`E2EContainer`): connect with a JoinResponse whose `room.sid` is empty, call `getSid()`, then deliver a RoomUpdate carrying the sid — the future never resolves. Test included in the linked PR; it times out on current `main` and passes with the fix.

## Suggested fix (PR attached)

Listen on the emitter that actually carries the event, re-check `_roomInfo` after registering (to close the race with `_applyRoomUpdate`), and cancel the listener when the future completes:

```dart
final cancelRoomUpdateListen =
    engine.signalClient.events.on<SignalRoomUpdateEvent>((event) {
  if (event.room.sid.isNotEmpty && !completer.isCompleted) {
    completer.complete(event.room.sid);
  }
});
...
return completer.future.whenComplete(cancelRoomUpdateListen);
```

## Related hardening suggestion

Even with `getSid()` fixed, `livekit_noise_filter.onPublish` awaiting an unbounded server-dependent future inside the publish critical section (and inside `SerialRunner`) means any slow sid turns into a frozen publish pipeline. Making the room-context handoff non-blocking there would remove the failure mode entirely — happy to file that separately if useful.

## Environment

- livekit_client: 2.8.1 (bug verified unchanged on `main` @ 9e05b79)
- livekit_noise_filter: 0.2.1
- Server: LiveKit Cloud
- Observed on macOS (native), but the affected code is pure Dart — all platforms

bkali:
Fixes #1151

## What

`Room.getSid()` waited for `SignalRoomUpdateEvent` on the Room's event emitter, but that event is only ever emitted on the signal client's emitter (and consumed by `_setUpSignalListeners`) — so when the server issues the room sid asynchronously (empty `room.sid` in the JoinResponse, real sid in a later RoomUpdate), the returned future never completed.

This deadlocks any `await room.getSid()` during the post-join window. Most visibly: `livekit_noise_filter`'s `onPublish` awaits it while the SDK holds the `SerialRunner` publish lock inside `_publishAudioTrack` — the mic track publishes and audio flows, but `setMicrophoneEnabled(true)` never returns and **every subsequent mute/unmute queues forever behind the lock** (observed in production on macOS with LiveKit Cloud; full analysis in #1151).

## How

- Listen on `engine.signalClient.events`, where `SignalRoomUpdateEvent` actually fires.
- Re-check `_roomInfo` after registering the listener, closing the race with `_applyRoomUpdate` (update landing between the early-return check and listener setup).
- Cancel the listener when the future completes (the previous listener was never removed).

## Tests

New `test/core/room_sid_test.dart` (using the existing `E2EContainer` mock harness):

1. sid present in the JoinResponse → `getSid()` returns immediately (guard, passed before).
2. sid empty in the JoinResponse, issued via a later RoomUpdate while `getSid()` is pending → **times out on `main`, passes with this fix**.
3. caller arriving after the RoomUpdate landed → returns the cached sid (guard, passed before).

Full test suite green (373 tests), `flutter analyze` clean on the changed files.

### DIFF
--- a/.changes/room-getsid-wrong-emitter
+++ b/.changes/room-getsid-wrong-emitter
@@ -1,0 +1,1 @@
+patch type="fixed" "Room.getSid() now resolves when the room sid is issued after the join response"
--- a/lib/src/core/room.dart
+++ b/lib/src/core/room.dart
@@ -1107,14 +1107,13 @@ extension RoomPrivateMethods on Room {
     // consumed by _setUpSignalListeners) — it never appears on the Room's
     // [events], so listen where it actually fires or the future returned
     // here never completes.
-    final cancelRoomUpdateListen =
-        engine.signalClient.events.on<SignalRoomUpdateEvent>((event) {
+    final cancelRoomUpdateListen = engine.signalClient.events.on<SignalRoomUpdateEvent>((event) {
       if (event.room.sid.isNotEmpty && !completer.isCompleted) {
         completer.complete(event.room.sid);
       }
     });
 
-    events.once<RoomDisconnectedEvent>((event) {
+    final cancelDisconnectListen = events.once<RoomDisconnectedEvent>((event) {
       if (!completer.isCompleted) {
         completer.complete('');
       }
@@ -1126,7 +1125,10 @@ extension RoomPrivateMethods on Room {
       completer.complete(_roomInfo!.sid);
     }
 
-    return completer.future.whenComplete(cancelRoomUpdateListen);
+    return completer.future.whenComplete(() {
+      cancelRoomUpdateListen();
+      cancelDisconnectListen?.call();
+    });
   }
 }
 
--- a/lib/src/core/room.dart
+++ b/lib/src/core/room.dart
@@ -1125,9 +1125,9 @@ extension RoomPrivateMethods on Room {
       completer.complete(_roomInfo!.sid);
     }
 
-    return completer.future.whenComplete(() {
-      cancelRoomUpdateListen();
-      cancelDisconnectListen?.call();
+    return completer.future.whenComplete(() async {
+      await cancelRoomUpdateListen();
+      await cancelDisconnectListen?.call();
     });
   }
 }
--- a/lib/src/core/room.dart
+++ b/lib/src/core/room.dart
@@ -1106,14 +1106,25 @@ extension RoomPrivateMethods on Room {
     // SignalRoomUpdateEvent is emitted on the signal client's emitter (and
     // consumed by _setUpSignalListeners) — it never appears on the Room's
     // [events], so listen where it actually fires or the future returned
-    // here never completes.
-    final cancelRoomUpdateListen = engine.signalClient.events.on<SignalRoomUpdateEvent>((event) {
+    // here never completes. Created via createListener() so it is cancelled
+    // with its owner.
+    final roomUpdateListener = engine.signalClient.createListener();
+    roomUpdateListener.on<SignalRoomUpdateEvent>((event) {
       if (event.room.sid.isNotEmpty && !completer.isCompleted) {
         completer.complete(event.room.sid);
       }
     });
 
     final cancelDisconnectListen = events.once<RoomDisconnectedEvent>((event) {
+      if (!completer.isCompleted) {
+        completer.complete('');
+      }
+    });
+
+    // Disposal without a disconnect event (Room.dispose during teardown)
+    // cancels the listeners above — complete instead of leaving the returned
+    // future pending forever.
+    onDispose(() async {
       if (!completer.isCompleted) {
         completer.complete('');
       }
@@ -1126,7 +1137,7 @@ extension RoomPrivateMethods on Room {
     }
 
     return completer.future.whenComplete(() async {
-      await cancelRoomUpdateListen();
+      await roomUpdateListener.dispose();
       await cancelDisconnectListen?.call();
     });
   }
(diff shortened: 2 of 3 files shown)


bkali:
Addressed both review findings in ae0a4c6:
- The `RoomDisconnectedEvent` once-listener's cancel func is now invoked in the same `whenComplete` as the signal listener's, so `getSid()` calls resolving via the sid path no longer accumulate listeners.
- Added the `.changes/room-getsid-wrong-emitter` release-notes entry (`patch type="fixed"`).

Also ran `dart format` to fix the failing format check. Full test suite green locally.

bkali:
Addressed in 8ca8f90 — `CancelListenFunc` returns a `Future`, so the `whenComplete` callback is now `async` and awaits both cancellations instead of discarding them. (Small note: the repo's `analysis_options.yaml` uses `lints/recommended`, which doesn't enable `unawaited_futures`/`discarded_futures` — analyze was green either way — but sequencing the cleanup is the right call regardless.)

bkali:
Addressed in 81ac48d. Worth noting the dispose-path hang was pre-existing (the old listener never fired at all, so the future hung in strictly more cases), but this PR is the right place to close it: an `onDispose` hook now completes the pending future with `''`, and the RoomUpdate listener is created via `createListener()` per the ownership convention in AGENTS.md. Added a dispose-while-waiting regression test — it times out without the guard, passes with it. Full suite green (374).

bkali:
Addressed in the latest commit — pending `getSid()` completers are now tracked in a set drained by the constructor's single dispose routine (each waiter removes itself on completion), so repeated calls register no per-call `onDispose` closures. Added a regression test asserting `disposeFuncCount` stays flat across repeated calls. Full suite green (375).

bkali:
Addressed in 22c1895 — an engine listener now completes the waiter directly from the JoinResponse (`EngineJoinResponseEvent`), covering callers who ask while the connection is still being established; disposed alongside the other listeners in `whenComplete`. Regression test added (times out without the listener). Full suite green (376).

With this, every completion path for `getSid()` is covered and tested: sid cached, sid via RoomUpdate, sid via JoinResponse mid-connect, disconnect, and disposal — happy to split anything further into a follow-up issue to keep this PR reviewable.

hiroshihorie (approved):
Approving. Verified locally, the new tests time out on `main` and all pass here.

Two notes. The `createListener` "cancelled with its owner" comment is inaccurate, but that wording came from AGENTS.md line 52, which is wrong, so I'll fix the doc and take the comment with it. And please do file your hardening suggestion for the processor await inside the publish lock, that is the part that turns any slow future into a mic the user cannot mute.

PR #1160, opened 2026-08-07
Integrate initial uniffi rust core

1egoman:
Some work was recently done on the rust sdks repo to get a dart uniffi bindgen integrated into the `livekit-uniffi` crate - here's a pull request: https://github.com/livekit/rust-sdks/pull/1183.

This is the other half of the change to integrate this these bindings into the flutter sdk. What this consisted of at a high level:
- Fixed some dart bindgen bugs that were discovered when building the latest `livekit-uniffi` release containing data streams and data tracks. 
  - https://github.com/Uniffi-Dart/uniffi-dart/pull/150
  - https://github.com/Uniffi-Dart/uniffi-dart/pull/154
  - https://github.com/Uniffi-Dart/uniffi-dart/pull/156
- Flutter native assets require at a minimum Flutter 3.38. So, I updated from flutter `>=3.27.0` to `>=3.38.0` / Dart `>=3.10.0`.
  - This is a significant version bump from what I can tell and some thorough testing should be done to ensure nothing was broken.
  - Also as part of this, the output of `dart format` changed to "tall style" ([more info](https://github.com/dart-lang/dart_style/issues/1253)) - and as part of this, in a9ef3a656663e826e68ce069f14ae4f837604f47, I've been forced to reformat a lot of files. **This was done mechanically by dart format, it shouldn't require an in depth review**
- Added in a facade layer to switch between native code (where the `livekit-uniffi` lib can be loaded successfully) and `web` where it cannot be.
  - The facade exposes `LiveKitUniffi.isAvailable` which can be used to guard uniffi code for future features
  - I've also taken the liberty of exposing the `LiveKitUniffi.buildVersion` function initially. This is a simple, sync call which returns a string and should be a good proof of concept that everything is wired up properly 

This change is a prerequisite to adding data streams v2 support to flutter.

Also note that this right now only builds locally. So CI fails and that is expected. Once all TODO items are complete, ci should build.

> [!WARNING]
> This pull request was largely LLM generated and while it has been reviewed by a human, that human is not a flutter expert. I have tested this and can confirm it works in the happy path, but no other validation has been done.
> 
> A more thorough review of this needs to occur from a domain expert before it could be merged.

## TODO
- [x] Get remaining pull requests for uniffi-dart bindgen submitted to fix discovered bugs
- [x] Get pull requests to upstream uniffi-dart package merged.
  - If this doesn't happen in a timely manor, consider potentially migrating the personal fork I made to an official livekit maintained fork before merging this, or figuring out some other way around the problem (like a patching script that runs post bindgen, etc).
- [x] Get the `livekit-uniffi` dart bindings published to https://pub.dev/, and cut over to the production build when making production flutter sdk releases.
  - Right now, this pull request assumes they are all locally built and located in `../rust-sdks/livekit-uniffi/packages/dart`.


lib/src/agent/chat/transcription_stream_receiver.dart
```diff
@@ -217,8 +217,9 @@ class TranscriptionStreamReceiver implements MessageReceiver {
     final displayTimestamp = partial?.timestamp ?? timestamp;
     final isLocalParticipant = _room.localParticipant?.identity == participantIdentity;
 
-    final ReceivedMessageContent content =
-        isLocalParticipant ? UserTranscript(displayContent) : AgentTranscript(displayContent);
+    final ReceivedMessageContent content = isLocalParticipant
+        ? UserTranscript(displayContent)
+        : AgentTranscript(displayContent);
 
     return ReceivedMessage(
```
1egoman, line 224: Image

Per the comment in the pull request body: the vast majority of this change is mechanical and can be ignored. In practice, it's more like a couple hundred lines of interesting stuff. Here's an example of the reformatting noise which is scoped to 5074c77.

lib/src/uniffi/uniffi.dart
```diff
@@ -0,0 +1,45 @@
...
+  /// Version string reported by the Rust core.
+  ///
+  /// The simplest possible round trip -- a synchronous, argument-free call
+  /// returning a string -- so it doubles as the smoke test that the whole
+  /// chain is wired up: build hook resolved the library, `@Native` bound the
+  /// symbol, and a value came back across the FFI boundary.
+  ///
+  /// Throws [UnsupportedError] on web.
+  static String get buildVersion => impl.buildVersion();
+}
```
1egoman, line 45: Here's the newly introduced `LiveKitUniffi` facade. This allows all these bits of functionality to be turned into noops on web.

pubspec.yaml
```diff
@@ -55,6 +57,17 @@ dependencies:
...
+  # Native platforms only — see lib/src/uniffi/.
+  livekit_uniffi: ^0.1.7
+
+# livekit_uniffi is not published to pub.dev yet, so it resolves out of a sibling
+# rust-sdks checkout produced by `cargo make dart-package`. See AGENTS.md. Drop
+# this once the package is published.
+dependency_overrides:
+  livekit_uniffi:
+    path: ../rust-sdks/livekit-uniffi/packages/dart
+
```
1egoman: The version bump and some local packaging machinery.

Calling out again because it is very important: this **CANNOT** be merged in its current state until the flutter bindings are distributed somehow. From what I can tell looking at the rust half, it looks like what was being planned was to publish them to https://pub.dev.

lib/src/track/local/audio.dart
```diff
@@ -69,10 +69,12 @@ class LocalAudioTrack extends LocalTrack with AudioTrack, LocalAudioManagementMi
     _throwIfAudioProcessingFailed(response);
 
     currentOptions = nextOptions;
-    events.emit(LocalTrackOptionsUpdatedEvent(
```
cloudwebrtc: It seems many of the differences are due to variations in Dart formatting. Could you please restore these formatting settings and just commit the code differences related to the PR?

cloudwebrtc (changes requested):

lib/src/track/local/audio.dart
cloudwebrtc: @hiroshihorie, could the inconsistency in formatting style be due to the different versions of dart-lang being used?  or it might just require updating `flutter-sdk` to the latest version (3.44.9) of the stable channel.

hiroshihorie: I think the formatter changed in the new dart SDK (This one), so I think not a problem, i will reformat all code after Dart SDK bump 🙂

lib/src/audio/audio_frame_capture.dart
```diff
@@ -21,7 +21,8 @@ import 'audio_frame_capture_native.dart' if (dart.library.js_interop) 'audio_fra
 /// PCM sample format for audio frame capture.
 enum AudioFormat {
   Int16('int16'),
-  Float32('float32');
+  Float32('float32')
+  ;
```
xianshijing-lk: curiously, what these format changes coming from ? 

are they from some tools ? 

I wonder if we can move these format refactoring to a separate PR instead

lib/src/track/local/audio.dart
1egoman: I called this out in the pull request body:

> Also as part of this, the output of dart format changed to "tall style" (https://github.com/dart-lang/dart_style/issues/1253) - and as part of this, in https://github.com/livekit/client-sdk-flutter/commit/a9ef3a656663e826e68ce069f14ae4f837604f47, I've been forced to reformat a lot of files. This was done mechanically by dart format, it shouldn't require an in depth review

and here:

https://github.com/livekit/client-sdk-flutter/pull/1160#discussion_r3738474559

TL;DR though - as hiroshi said, this formatting is due to the dart version upgrade which is required for flutter native asset support, they seemed to have changed their code formatting style. All formatting changes were isolated into 5074c77 to try to make the review easier.

lib/src/audio/audio_frame_capture.dart
1egoman: https://github.com/livekit/client-sdk-flutter/pull/1160#discussion_r3749765038

I could potentially isolate these code formatting changes into their own PR, but if so, then CI would not pass because of the change to what `dart format` validates. They are segmented into their own commit so I can do whatever is easiest here.

hiroshihorie (approved):

PR #1169, opened 2026-08-10
Ensure roomOptions in Room.connect(...) are using the locally defined variable, NOT implicitly this.roomOptions

1egoman:
While working on #1166, I discovered this bug incidentally, so here is an isolated fix.

`connect`'s first line is `var roomOptions = this.roomOptions;`. However, a parameter of the same name (`roomOptions`) was in scope. After some unexpected behavior, I realized that dart permits a local to shadow a parameter (silently, with no warning and the local winning) so every `RoomOptions` a caller passed to `connect` was discarded and the Room's own options were used instead. Nothing surfaced the mismatch.

The `roomOptions` parameter on `connect` seems to be deprecated in favor of the parameter on the `Room` constructor, but deprecated is not the same as inert: while it is still accepted it has to take effect. It now does, falling back to the Room's options when absent. `Engine.connect` already adopted whatever it was handed, so the value propagates from there without further changes.

The local is renamed to `effectiveRoomOptions`, since restoring the parameter's visibility is the whole point and leaving two things called `roomOptions` in one scope is what caused this.

I also addeds a regression test, and threaded `connectOptions`/`roomOptions` through the E2E container so it can be exercised. I verified the test failed against the old shadowing behavior.

### DIFF
--- a/lib/src/core/room.dart
+++ b/lib/src/core/room.dart
@@ -274,21 +274,26 @@ class Room extends DisposableChangeNotifier with EventsEmittable<RoomEvent> {
     @Deprecated('deprecated, please use roomOptions in Room constructor') RoomOptions? roomOptions,
     FastConnectOptions? fastConnectOptions,
   }) async {
-    var roomOptions = this.roomOptions;
-    if (lkPlatformIs(PlatformType.web) && (roomOptions.networkOptions.certificatePinning?.isEnabled ?? false)) {
+    var effectiveRoomOptions = roomOptions ?? this.roomOptions;
+    if (lkPlatformIs(PlatformType.web) &&
+        (effectiveRoomOptions.networkOptions.certificatePinning?.isEnabled ?? false)) {
       throw UnsupportedError('Certificate pinning is not supported on Flutter web, '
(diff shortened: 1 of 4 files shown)


lib/src/core/room.dart
```diff
@@ -274,21 +274,26 @@ class Room extends DisposableChangeNotifier with EventsEmittable<RoomEvent> {
     @Deprecated('deprecated, please use roomOptions in Room constructor') RoomOptions? roomOptions,
     FastConnectOptions? fastConnectOptions,
   }) async {
-    var roomOptions = this.roomOptions;
-    if (lkPlatformIs(PlatformType.web) && (roomOptions.networkOptions.certificatePinning?.isEnabled ?? false)) {
+    var effectiveRoomOptions = roomOptions ?? this.roomOptions;
```
xianshijing-lk, line 277: what is this.roomOptions ? is it a default roomOptions ? or it caches the latest roomOptions ? 

Can we rename it to be more clear and less error prone ?

hiroshihorie (approved):

cloudwebrtc (approved):

PR #1177, opened 2026-08-20
Fix use-after-free freeze in Linux TaskRunnerLinux::EnqueueTask

p-divita-salto:
On Linux, the app can freeze completely if an AudioRendererSink (or VisualizerSink) is torn down via stopAudioRenderer/stopVisualizer while an audio frame's task is still queued for dispatch on the GLib main loop.
The freeze is permanent and the whole UI thread hangs.

A gdb thread dump at the moment of freeze showed the main/GTK thread permanently blocked:

```
  #5 std::mutex::lock                                                                                                                                                                                                                                                        
  #7 livekit_client_plugin::TaskRunnerLinux::EnqueueTask(...)::$_0::operator()                                                                                                                                                                                               
     at linux/task_runner_linux.cc:19                                                                                                                                                                                                                                        
  #11 g_main_context_iteration
```

No other thread was holding the contended mutex — the signature of a stale pointer rather than live contention.

AudioRendererSink::PostEvent() calls TaskRunnerLinux::EnqueueTask(), which schedules a GLib idle callback via g_main_context_invoke(context, callback, this), capturing a raw TaskRunnerLinux*.

stopAudioRenderer destroys the owning sink (and its TaskRunnerLinux) synchronously:

it->second->RemoveSink();
renderers_.erase(it);

If an idle callback was already scheduled but had not yet run at that moment, the main loop later dispatches it against freed memory, locking a std::mutex inside deallocated memory is undefined behavior, which manifests here as a permanent futex wait.

My fix works as follow:

1. TaskRunnerLinux now inherits std::enable_shared_from_this and is held via std::shared_ptr (both AudioRendererSink and VisualizerSink share this ownership pattern and exposure).
2. EnqueueTask passes a heap-allocated std::weak_ptr<TaskRunnerLinux> through g_main_context_invoke_full (instead of a raw pointer through g_main_context_invoke), freeing it via the accompanying GDestroyNotify.
3. The dispatched callback locks the weak pointer and no-ops if the runner has already been destroyed, instead of touching freed memory.

I reproduced and verified fixed on Linux: start a call, enable an audio renderer (I have a captions/transcription use case), talk for a while so several frames are queued, then tear down the renderer and end the call: no longer freezes. 

I ran repeatedly without reproduction after the fix; reproduced reliably before it


cloudwebrtc (approved):

hiroshihorie (approved):
Thanks for the thorough diagnosis, the gdb trace made this easy to follow.

The fix is the right shape: a weak_ptr through g_main_context_invoke_full with a destroy notify is the idiomatic GLib pattern, and the reason the old code was safe in flutter-webrtc (one runner living for the whole process) stopped holding once each sink got its own runner that is destroyed synchronously in stopAudioRenderer / stopVisualizer.

I pushed two follow-up commits to the branch:
- merged main so the Dart Analyze / Dart Format / Android failures (all pre-existing on the base commit) go away
- run the queued tasks outside tasks_mutex_ so a task that re-enters EnqueueTask cannot deadlock on the non-recursive mutex, plus a comment on why weak_from_this() is safe from the audio thread (RemoveSink blocks until in-flight OnData returns) and some comment tidying

I also verified the fix in isolation: a small GLib harness that drives TaskRunnerLinux the way AudioRendererSink does (enqueue from a second thread, destroy the runner, then drain the main context) under AddressSanitizer. Pre-PR code fails 20/20 runs on the stale mutex, this branch passes 20/20 with zero tasks run after destroy.

Issue #1042, opened 2026-03-26
[bug] iOS withPreConnectAudio throws an error when connecting to room


adambeer:
**Describe the bug**
When connecting to a livekit room using withPreConnectAudio on iOS, the following error is produced:

Unable to start local recording: Error: adm api failed with code: -1

#0      NativeAudioManagement.startLocalRecording (package:flutter_webrtc/src/native/audio_management.dart:79:9)
<asynchronous suspension>
#1      PreConnectAudioBuffer.startRecording (package:livekit_client/src/preconnect/pre_connect_audio_buffer.dart:149:7)
<asynchronous suspension>
#2      RoomPreConnect.withPreConnectAudio (package:livekit_client/src/core/room_preconnect.dart:53:5)
<asynchronous suspension>

**To Reproduce**

Use `withPreconnectAudio` like in the docs: https://docs.livekit.io/agents/multimodality/audio/ on an iOS device and try to connect to a room.

**Expected behavior**

**Platform information**

* **Flutter version**: 3.41.4
* **Plugin version**: 2.7.0
* **Flutter target OS**: iOS real device
* **Flutter target OS version**: iOS 17.5.1

hiroshihorie:
I think something in our dependency (flutter_webrtc) changed, investigating this one...

hiroshihorie:
This should be fixed in 2.12.0 (#1182): the audio session is now configured from engine state before recording starts, so `withPreConnectAudio` no longer fails on iOS. Please reopen if you still hit it after upgrading. Thanks!

PR #1182, opened 2026-08-27
Resolve a default audio session from engine state when no policy was pushed

### LINKED ISSUE #1165: [bug] Audio engine returned error code
**Describe the bug**

On the pre join screen, one of our users reported this bug:

```
AudioProcessingException: AudioProcessingException(applyFailed): Audio engine returned error code: -9001
#0      LocalAudioTrack.startCapture (package:livekit_client/src/track/local/audio.dart:89)
  File "<asynchronous suspension>"
#2      Track.start (package:livekit_client/src/track/track.dart:113)
  File "<asynchronous suspension>"
#4      _PrejoinSessionScreenState._initializeLocalAudio (package:totem_core/features/sessions/screens/loading_screen.dart:224)
```

**To Reproduce**

Try to create a LocalAudioTrack. See bug.

**Expected behavior**

No bug and audio track is initialized correctly.

**Platform information**

* **Flutter version**:  Flutter (Channel stable, 3.44.4, on macOS 26.5.2 25F84 darwin-arm64, locale pt-BR)
* **Plugin version**: v2.9.0
* **Flutter target OS**:  iOS
* **Flutter target OS version**: Flutter 3.44.6 (stable)

**Additional Information**

* https://github.com/livekit/client-sdk-swift/issues/849

hiroshihorie:
## Problem

On iOS the WebRTC audio engine refuses to enable recording unless the audio session category permits input. Its pre-enable check returns `kAudioEngineErrorAudioSessionInvalidCategory` (-9001), which the SDK reported as `AudioProcessingException(applyFailed): Audio engine returned error code: -9001`.

`livekit_client` owns the iOS audio session since #1108 (`LiveKitPlugin.swift` disables flutter_webrtc's session management at registration). The native engine observer (`LKAudioEngineObserver.willEnableEngine`) is the right hook and runs before the engine's check, but it only applied a configuration that Dart had pushed. In automatic mode the only push site was `Room.connect`, so anything that started recording earlier met an empty cache, the observer returned "proceed" with the session still `soloAmbient`, and the engine rolled back:

- `Room.withPreConnectAudio`, which `SessionOptions.preConnectAudio` enables by default, so the Flutter agent starter failed on every "Start call"
- a pre-join microphone preview (#1165)
- an engine start driven from native before the Flutter side exists, for example the plugin's static `setEngineAvailability` on a CallKit killed-state wake

Because the preconnect throw happens before `connect`, the cache was never seeded and the failure repeated on every attempt.

## How the Swift SDK handles this

`AudioSessionEngineObserver.engineWillEnable` derives the session configuration from the requested engine state alone (`playAndRecord` presets while recording, `playback` for playout only), synchronously inside the engine's enable call. Nothing is configured "before connect". The engine asks, the observer configures, the engine starts. The Flutter plugin already has the same observer in the same place, it just had no built-in policy.

## Fix

`LKAudioEngineObserver.effectiveConfigurationLocked` now resolves a built-in `playAndRecord` preset (`allowBluetooth | allowBluetoothA2DP | allowAirPlay`, `videoChat`) whenever nothing has been pushed and automatic management is on. The existing playout-only `playback` branch applies to it as well. Manual mode still leaves the session alone. The Dart-pushed policy becomes an override rather than a prerequisite, and for the default `AudioSessionOptions.communication` it pushes the same values, so the connect-time push does not change the live session.

The preset is built on a copy of the shared `RTCAudioSessionConfiguration.webRTC()` object, and it is best-effort: if applying it fails, the engine start proceeds and the ADM's own pre-enable checks still gate recording, so apps whose own session was already valid are not newly rolled back with -4100. Only a policy the app actually pushed keeps failing the engine start hard. The mode is fixed to `videoChat` because it matches the Dart default speaker preference, and a non-default preference always arrives as a pushed policy whose mode already carries it, so the preset can never observe anything else.

`LocalAudioTrack.startCapture` additionally pushes the resolved Dart policy to native before recording starts (cache-only while the engine is idle in automatic mode). Flutter-driven starts therefore always use the real Dart policy, and the built-in preset only stands in for engine starts that happen before the Flutter side exists, for example the plugin's static `setEngineAvailability` on a CallKit killed-state wake.

The engine observer is also shared across plugin registrations now: a second Flutter engine registering in the same process (add-to-app, `FlutterEngineGroup`) previously reset the pushed policy and management mode silently, which the preset would have turned into an unwanted session activation. Only the notification channel is rebound per registration.

## Error mapping

Audio device module results now get their own error codes on the `startLocalRecording`, `setEngineAvailability`, `setMicrophoneMuteMode` and `stopLocalRecording` channels, mirroring client-sdk-swift's `checkAdmResult`. A mute-mode change can rebuild the engine and hit the same pre-enable checks as a recording start, so all entry points surface the same failure the same way:

| ADM result | Native error code | Dart exception |
| --- | --- | --- |
| -9000 `InsufficientDevicePermission` | `deviceAccessDenied` | `TrackCreateException` |
| -9001 `AudioSessionInvalidCategory` | `audioSessionInvalidCategory` | `AudioSessionException` (new) |
| -4100 `FailedToConfigureAudioSession` | `audioSessionConfigureFailed` | `AudioSessionException` (new) |
| anything else | caller fallback (`applyFailed`, `setEngineAvailability`, ...) | unchanged |

`AudioSessionException` is a new public class with its own changeset entry.

## Relation to #1179

@MaxHeimbrock's #1179 diagnosed this first and fixes it by pushing the policy from `LocalAudioTrack.startCapture` (`prepareRecording()`). This PR ends up covering both layers: `startCapture` pushes the resolved policy like #1179 does (without a Dart-side flag tracking native cache state), and the native observer is additionally self-sufficient like Swift's, so engine enables that never pass through Dart are covered too. With this merged, `prepareRecording()` becomes redundant.

## Testing

- Flutter agent starter on an iPhone 17 Pro (iOS 27.0): unpatched, `Start call` failed with -9001 on every attempt. Patched, `startCapture()` succeeds two seconds before the connect-time policy push, the preconnect buffer is sent to the agent, and the call connects.
- `flutter analyze`, `flutter test`, `dart format --set-exit-if-changed`, `import_sorter --exit-if-changed` clean.
- New unit tests cover the error mapping. The iOS-gated branches are not reachable from unit tests (`lkPlatform()` has no seam), same limitation as #1179 noted.

Fixes #1165
Refs #1042

### DIFF
--- a/.changes/audio-engine-error-mapping
+++ b/.changes/audio-engine-error-mapping
@@ -1,0 +1,1 @@
+patch type="changed" "Microphone permission and audio session failures from the audio engine surface as TrackCreateException and the new AudioSessionException instead of AudioProcessingException(applyFailed)"
--- a/.changes/audio-engine-error-mapping
+++ b/.changes/audio-engine-error-mapping
@@ -1,1 +1,1 @@
-patch type="changed" "Microphone permission and audio session failures from the audio engine surface as TrackCreateException and the new AudioSessionException instead of AudioProcessingException(applyFailed)"
+patch type="changed" "Microphone permission and audio session failures now throw TrackCreateException / AudioSessionException instead of AudioProcessingException"
--- a/.changes/audio-session-exception
+++ b/.changes/audio-session-exception
@@ -1,1 +1,1 @@
-minor type="added" "Add AudioSessionException, thrown when the Apple audio session cannot be configured for or does not permit the requested audio operation"
+minor type="added" "Add AudioSessionException for iOS audio session failures"
--- a/.changes/audio-session-exception
+++ b/.changes/audio-session-exception
@@ -1,1 +1,1 @@
-minor type="added" "Add AudioSessionException for iOS audio session failures"
+patch type="added" "Add AudioSessionException for iOS audio session failures"
--- a/.changes/audio-session-exception
+++ b/.changes/audio-session-exception
@@ -1,0 +1,1 @@
+minor type="added" "Add AudioSessionException, thrown when the Apple audio session cannot be configured for or does not permit the requested audio operation"
--- a/.changes/native-audio-session-preset
+++ b/.changes/native-audio-session-preset
@@ -1,0 +1,1 @@
+patch type="fixed" "iOS: the audio engine now configures a playAndRecord audio session on its own when recording starts before any session policy was pushed (pre-connect audio, pre-join microphone preview, CallKit-driven engine start), instead of failing with audio engine error -9001"
--- a/.changes/native-audio-session-preset
+++ b/.changes/native-audio-session-preset
@@ -1,1 +1,1 @@
-patch type="fixed" "iOS: the audio engine now configures a playAndRecord audio session on its own when recording starts before any session policy was pushed (pre-connect audio, pre-join microphone preview, CallKit-driven engine start), instead of failing with audio engine error -9001"
+patch type="fixed" "iOS: audio session is configured for recording even when capture starts before connect (pre-connect audio, pre-join mic), fixing audio engine error -9001"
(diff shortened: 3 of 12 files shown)


cloudwebrtc (approved):

MaxHeimbrock:
I am able to confirm this also fixes the issue 👍

MaxHeimbrock (approved):

Issue #553, opened 2024-07-11
[bug] Actions stop working after second connect attempt to same room flutter livekit client


RoyMatipa:
**Describe the bug**

Hello! When I try to disconnect from a room in my Flutter app, I've noticed that it works the first time, but if you connect again and then try to disconnect a second time, it doesn't do anything and no error is returned. Additionally, other actions like `setCameraEnabled` and `setMicrophoneEnabled` also stop working after the second connection to the same room.

**To Reproduce**

1. Connect to a room.
2. Disconnect from the room.
3. Connect to the same room.
4. Disconnect from the room (This step doesn't work for some reason).
5. Try to use `setCameraEnabled` or `setMicrophoneEnabled` (These actions also stop working SOMETIMES).

**Expected behaviour**

The app should successfully disconnect from the room on the second attempt as it does on the first attempt, and actions like `setCameraEnabled` and `setMicrophoneEnabled` should continue to work.

**Platform information**

* **Flutter version**: 3.22.1
* **Plugin version**: livekit_client: ^2.2.0
* **Flutter target OS**: Android/iOS/Web (Only tested on Android Emulator running android 14 & Samsung Galaxy S22+ running Android 14)
* **Flutter target OS version**: Android 14
* **Flutter console log**:

DumbDev168:
I have the same issue in version 2.2.0. 
In version 2.1.6 it's working fine.

PR #1208, opened 2026-09-15
fix: disconnect() tears down locally after sending the Leave instead of waiting for the server echo

hiroshihorie:
## Problem

`engine.disconnect()` sent the Leave and relied on the server's echo to run cleanup and emit `EngineDisconnectedEvent`, and `Room.disconnect()` waited up to 10 s for that event. Media nodes drop queued leave messages when they close the signal sink (`pkg/routing/signal.go`, `signalMessageSink.Close`, comment in place), so the echo can be lost. Every lost echo was a 10 s stall, a `TimeoutException` out of `disconnect()`, and a room left half torn down because `Room.disconnect()` never reached its own cleanup.

Found by the e2e-flutter suite (livekit/e2e-flutter#7): one agent per run against a local livekit-server 1.13.6, `SignalClient did disconnect` with no `Leave received` before it while the server logged `sendLeave: true`.

This design dates from the Dec 2023 engine rewrite (v1.5.5). #559 (v2.2.2, issue #553) made `Room.disconnect()` wait for the event, which turned a silently missing event into a visible stall. No other SDK waits on the server here: JS, Swift, Android and Rust all send the Leave and tear down locally.

## Fix

Flutter now does the same.

- `engine.disconnect()` sends the Leave, runs `cleanUp()` and emits `EngineDisconnectedEvent` in one go. No dependency on the echo.
- A Leave that arrives after `disconnect()` started is the echo, or a server Leave racing ours, and is ignored.
- `Room.disconnect()` awaits the engine teardown and then its own cleanup, no wait on the server and no timeout. The engine event handler skips its cleanup while `Room.disconnect()` is running, so teardown has one owner and a `connect()` right after `disconnect()` returns cannot race a straggling handler. That is the case behind #553, now covered by a test.

The first commit on this branch was a narrower fallback that completed the disconnect on socket close while keeping the echo dependency. The second commit replaces it with the alignment above.

## Tests

`test/core/disconnect_event_test.dart`, four added:
- socket close without echo completes `disconnect()` with one `clientInitiated` event
- echo followed by socket close emits exactly one event
- `disconnect()` returns with engine and room torn down when the server sends nothing at all
- connect again to the same room right after `disconnect()` returns

The last two fail on main. Full suite 428 passing.

## Not yet run

The e2e-flutter suite against this branch. The fallback version of the branch turned the suite's one warning into zero and cut the run by 9 s. Xcode is not available on my machine at the moment, so the final version has not had its real-server run yet. Will report here when it has.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


hiroshihorie:
e2e-flutter suite (livekit/e2e-flutter#7) against this branch, local livekit-server 1.13.6, macOS, three full runs on 2026-09-16:

- run 1: 8 pass, 10 skip, 0 warn, 0 fail
- run 2: 8 pass, 10 skip, 0 warn, 1 fail in ParticipantMetadataTest, a race in the test itself (un-awaited waitUntil), fixed in e2e-flutter 920b3b2
- run 3: 8 pass, 10 skip, 0 warn, 0 fail

ParticipantAttributesTest, the scenario whose agent hit the lost Leave echo on every run before, completed its teardown in about 0.5 s on all three runs with no warning. The 10 skips are camera scenarios, not runnable on this machine.

